PR #2, opened 2021-04-21
Add prestop for graceful shutdown envoy pods

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-8790

Add `preStop` for graceful shutdown.
ref: https://github.com/scalar-labs/scalar-envoy/pull/10

# Done
- envoy
  - Add `post helthcheck`
  - Add `sleep`

### DIFF
--- a/charts/scalardl/templates/ledger/deployment.yaml
+++ b/charts/scalardl/templates/ledger/deployment.yaml
@@ -74,6 +74,10 @@ spec:
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "sleep 30"]
           resources:
             {{- toYaml .Values.ledger.resources | nindent 12 }}
       {{- with .Values.ledger.nodeSelector }}
--- a/charts/scalardl/templates/ledger/deployment.yaml
+++ b/charts/scalardl/templates/ledger/deployment.yaml
@@ -74,10 +74,6 @@ spec:
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1
-          lifecycle:
-            preStop:
-              exec:
-                command: ["/bin/sh", "-c", "sleep 30"]
           resources:
             {{- toYaml .Values.ledger.resources | nindent 12 }}
       {{- with .Values.ledger.nodeSelector }}
--- a/charts/scalardl/Chart.yaml
+++ b/charts/scalardl/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: scalardl
 description: Implementation scalardl.
 type: application
-version: 1.2.2
+version: 1.2.3
 appVersion: 2.1.0
 deprecated: false
 icon: https://scalar-labs.com/wp-content/themes/scalar/assets/img/logo_scalar.svg
--- a/charts/scalardl/README.md
+++ b/charts/scalardl/README.md
@@ -1,7 +1,7 @@
 # scalardl
 
 Implementation scalardl.
-Current chart version is `1.2.2`
+Current chart version is `1.2.3`
 
 ## Values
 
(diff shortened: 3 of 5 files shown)


charts/scalardl/templates/ledger/deployment.yaml
```diff
@@ -74,6 +74,10 @@ spec:
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "sleep 30"]
```
tei-k, line 80: Since `/bin/sh` doesn't forward signals to any child processes, the SIGTERM we sent never reached our script.
So the pods will always be deleted after `60` seconds. (`terminationGracePeriodSeconds: 60`)
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0   4276   796 ?        Ss   Apr19   0:00 /bin/sh -c dockerize -template ledger.properties.tmpl:ledger.properties     -template log4j.properties.tmpl:log4j.proper
root         7  0.0  0.0  10052  5340 ?        Sl   Apr19   0:00 dockerize -template ledger.properties.tmpl:ledger.properties -template log4j.properties.tmpl:log4j.properties ./bin/scal
root        12  0.1  3.4 3320860 281508 ?      Sl   Apr19   3:45 /usr/local/openjdk-8/bin/java -Djava.security.manager -Djava.security.policy=security.policy -Dlog4j.configuration=file:
```

superbrothers: Two problems occur together here. One is that `sh` doesn't propagate signals to its children, and the other is that it ignores SIGTERM because it's running with PID 1.

A process with PID 1 does not receive a signal unless it is explicitly handling the signal. This is called "PID 1 problem".

Since `dockerize` seems to handle the signals, we can change `dockerize` process to PID 1 as follows to make it work.

- https://github.com/jwilder/dockerize/blob/9ce2e80619962abe9ba90619e009d29943446ab3/exec.go#L30-L44

To set `dockerize` to PID 1, change the `Dockerfile` as follows:

```dockerfile
ENTRYPOINT ["dockerize"]
CMD ["-template", "ledger.properties.tmpl:ledger.properties", \
    "-template", "log4j.properties.tmpl:log4j.properties", \
    "./bin/scalar-ledger", "--config ledger.properties"]
```

Another solution is Pod `shareProcessNamespace` feature. This is a feature that allows you to share PIDs between containers in a pod, with the effect that the process does not run with PID 1. However, you need to modify your `Dockerfile` to handle the signal correctly, since it will be executed with `sh -c`.

feeblefakie: @superbrothers Thank you!

@tei-k 
Since we need to update the Dockerfile and check if the current ledger support SIGTERM gracefully to make it work as expected, let's focus on Envoy part for this PR and work on for Ledger in another PR.

feeblefakie (approved):
LGTM! (let's separate the Ledger changes)

charts/scalardl/templates/ledger/deployment.yaml
tei-k: @superbrothers Thanks!

>Since we need to update the Dockerfile and check if the current ledger support SIGTERM gracefully to make it work as expected, let's focus on Envoy part for this PR and work on for Ledger in another PR.

@feeblefakie Removed ledger part in fc9467f

ymorimo (approved):
LGTM!

feeblefakie:
@tei-k Can you fix the CI failure?

tei-k:
>Can you fix the CI failure?

@feeblefakie 
We need to create new image `1.0.1` for `scalar-envoy`. I can't push the image to `ghcr` with permission error.

tei-k:
@feeblefakie 

CI was succeeded after pushed the envoy image `scalar-envoy:1.0.1`.

superbrothers (approved):

PR #15, opened 2021-08-16
Update Chart.yaml

feeblefakie:
Updated `Chart.yaml`s since their descriptions are weird and the maintainers are not updated.


supl (approved):
I am not familiar with Helm Chart but the changes look good to me.

tei-k (approved):
LGTM! Thank you!

PR #33, opened 2021-09-30
Modify index yaml

miseyu:
# Summary
- sync index yaml(cr index command)
- `scalardl`, `scalardl-audit`, and `scalardb` charts have been added to the index yaml.
- The reason why I didn't run it with github actions is that github actions targets the prior commit contents, so it didn't deploy when I ran it, so I had to run the chart releaser manually..

# Detail
## Deployment details
```
$ cr package charts/scalardl
$ cr package charts/scalardb
$ cr package charts/scalardl-audit
$ cr upload --skip-existing
$ cr index --charts-repo https://scalar-labs.github.io/helm-charts --index-path index.yaml
```

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -1,6 +1,91 @@
 apiVersion: v1
 entries:
+  envoy:
+  - apiVersion: v2
+    appVersion: 1.1.0
+    created: "2021-09-17T15:44:08.01326+09:00"
+    description: Envoy Proxy for Scalar applications
+    digest: 1c8b4801539faef21bafa40ae01210562df0f911b07636042dbf5a9d44558dd3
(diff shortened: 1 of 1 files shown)


index.yaml
```diff
@@ -97,30 +215,4 @@ entries:
     urls:
     - https://github.com/scalar-labs/helm-charts/releases/download/schema-loading-1.3.0/schema-loading-1.3.0.tgz
     version: 1.3.0
-  envoy:
```
miseyu, line 100: The order has been switched from above. (This is because it was generated automatically.)

feeblefakie (approved):
LGTM! Thank you!

PR #35, opened 2021-10-05
Bump up scalar-envoy version to 1.2.0 for envoy charts

miseyu:
## Summary
- bump up scalar-envoy version to 1.2.0 (https://github.com/orgs/scalar-labs/packages/container/scalar-envoy/8462600?tag=1.2.0)
- add serviceListeners environment

### Verify that environment variables are applied.
``` yaml
envoyConfiguration:
  serviceListeners: ledger-service:40051,ledger-privileged:40052
```

* Envoy SDS API
``` sh
curl http://0.0.0.0:9001/listeners?format=json
{
 "listener_statuses": [
  {
   "name": "ledger-service",
   "local_address": {
    "socket_address": {
     "address": "0.0.0.0",
     "port_value": 40051
    }
   }
  },
  {
   "name": "ledger-privileged",
   "local_address": {
    "socket_address": {
     "address": "0.0.0.0",
     "port_value": 40052
    }
   }
  }
 ]
}

```

### DIFF
--- a/charts/envoy/values.yaml
+++ b/charts/envoy/values.yaml
@@ -4,12 +4,14 @@ replicaCount: 3
 envoyConfiguration:
   # envoyConfiguration.adminAccessLogPath -- admin log path
   adminAccessLogPath: /dev/stdout
+  # envoyConfiguration.serviceListeners -- list of service name and port
+  serviceListeners: scalar-service:50051,scalar-privileged:50052
 
 image:
   # image.repository -- Docker image
   repository: ghcr.io/scalar-labs/scalar-envoy
   # image.tag -- Docker tag
-  version: 1.1.0
+  version: 1.2.0
   # image.pullPolicy -- Specify a imagePullPolicy
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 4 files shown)


charts/envoy/values.yaml
```diff
@@ -4,12 +4,14 @@ replicaCount: 3
 envoyConfiguration:
   # envoyConfiguration.adminAccessLogPath -- admin log path
   adminAccessLogPath: /dev/stdout
+  # envoyConfiguration.serviceListeners -- list of service name and port
+  serviceListeners: scalar-service:50051,scalar-privileged:50052
```
feeblefakie, line 8: Just a confirmation.
This is not necessary to make it work, right?
(since `scalar-service:50051,scalar-privileged:50052` is the default string specified in envoy.yaml)

feeblefakie (approved):
LGTM! Thank you!
Left one confirmation question.

charts/envoy/values.yaml
miseyu: > This is not necessary to make it work, right?

Yes, it is not a required setting.
This is only written as an example. If that is the case, would it be better to leave this blank and provide an example in the comments?

feeblefakie: Thank you! I think it's fine! (I think README is updated by this so it's helpful for users.)

feeblefakie:
@ymorimo @superbrothers I'm merging it but PTAL when you get a chance.

superbrothers (commented):
LGTM :+1:

PR #38, opened 2021-10-12
Update the default value of ledgerProofEnabled to false

scalar-boney:
Changing the default value of `ledgerProofEnabled` to `false`. Otherwise, always need to create `ledger-keys` secret.


scalar-boney:
> @scalar-boney We found the cause of the problem and fixed it.🙇
> 
> [b97a10a](https://github.com/scalar-labs/helm-charts/commit/b97a10a8d2866da1da1e296f0f5ce1d2d0fb89f2)

Thank you very much!

PR #46, opened 2021-11-05
Modify index.yaml for scalardl 3.1.1

miseyu:
## Summary
- bumpup scalard app to 3.1.1
- I'm releasing this manually because I wanted to give a patch version of the previous major version, since the minor version was updated first!

### package sha256
```
$ sha256sum scalardl-audit-1.1.2.tgz
0e8866b20eb6f8dd109e5d88db70b74c9e35ec5a3cef662298e5677bcb9fb026

$ sha256sum scalardl-3.1.3.tgz
c7e7176c759f3f09c40720e72aac66f21de7c81bbed92749aac507d5be60b1a7
```

### scalardl diff
https://github.com/scalar-labs/helm-charts/compare/scalardl-3.1.2...scalardl-3.1.3

### scalardl-audit diff
https://github.com/scalar-labs/helm-charts/compare/scalardl-audit-1.1.1...scalardl-audit-1.1.2

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -218,6 +218,37 @@ entries:
     - https://github.com/scalar-labs/helm-charts/releases/download/scalardl-3.2.0/scalardl-3.2.0.tgz
     version: 3.2.0
   - apiVersion: v2
+    appVersion: 3.1.1
+    created: "2021-11-05T06:14:54.009954637Z"
+    dependencies:
+    - condition: envoy.enabled
+      name: envoy
+      repository: https://scalar-labs.github.io/helm-charts
+      version: ~1.1.1
+    description: Scalar DL is a tamper-evident and scalable distributed database.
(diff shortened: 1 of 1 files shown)


index.yaml
```diff
@@ -217,6 +217,37 @@ entries:
     urls:
     - https://github.com/scalar-labs/helm-charts/releases/download/scalardl-3.2.0/scalardl-3.2.0.tgz
     version: 3.2.0
+  - apiVersion: v2
```
feeblefakie, line 220: Do we manage this file manually?
Do we need a section for each version?
If so, I have some concerns since the file has many redundancies and it seems error-prone to update this file.

feeblefakie (commented):
Thank you!
Left one question.

index.yaml
miseyu: @feeblefakie 
Yes, you are right.
I think we need to think about how to upgrade the patch versions of the lower major and minor versions operationally.

PR #49, opened 2021-12-03
Fixed the items pointed out in the security audit in envoy charts

miseyu:
## Summary
- We used [kubeauidt](https://github.com/Shopify/kubeaudit) to conducted a security audit in envoy charts.
- I fixed the items pointed out in the security audit.
- I created pod security policy to envoy charts
- The reason for creating the `pod security policy` is to avoid violating the settings that comply with the security audit log.

## Security audit summary
``` sh
$ kubeaudit all -k kube-audit.yaml -f <(helm template scalr-envoy charts/envoy) -m error

-- [error] AutomountServiceAccountTokenTrueAndDefaultSA
   Message: Default service account with token mounted. automountServiceAccountToken should be set to 'false' on either the ServiceAccount or on the PodSpec or a non-default service account should be used.

-- [error] CapabilityOrSecurityContextMissing
   Message: Security Context not set. The Security Context should be specified and all Capabilities should be dropped by setting the Drop list to ALL.
   Metadata:
      Container: envoy

-- [error] RunAsNonRootPSCNilCSCNil
   Message: runAsNonRoot should be set to true or runAsUser should be set to a value > 0 either in the container SecurityContext or PodSecurityContext.
   Metadata:
      Container: envoy

-- [error] AllowPrivilegeEscalationNil
   Message: allowPrivilegeEscalation not set which allows privilege escalation. It should be set to 'false'.
   Metadata:
      Container: envoy

-- [error] SeccompAnnotationMissing
   Message: Seccomp annotation is missing. The annotation seccomp.security.alpha.kubernetes.io/pod: runtime/default should be added.
   Metadata:
      MissingAnnotation: seccomp.security.alpha.kubernetes.io/pod

```
* kube-auidt.yaml
``` kube-auidt.yaml
enabledAuditors:
  apparmor: false
  asat: true
  capabilities: true
  hostns: true
  image: true
  limits: true
  mounts: true
  netpols: true
  nonroot: true
  privesc: true
  privileged: true
  rootfs: false
  seccomp: true
auditors:
  capabilities:
    allowAddList: ['NET_BIND_SERVICE']
```

### DIFF
--- a/charts/envoy/templates/deployment.yaml
+++ b/charts/envoy/templates/deployment.yaml
@@ -16,6 +16,10 @@ spec:
   {{- end }}
   template:
     metadata:
+    {{- if .Values.podAnnotations }}
+      annotations:
+        {{- toYaml .Values.podAnnotations | nindent 8 }}
+    {{- end }}
       labels:
         {{- include "envoy.selectorLabels" . | nindent 8 }}
     spec:
@@ -26,6 +30,7 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       restartPolicy: Always
+      automountServiceAccountToken: false
       terminationGracePeriodSeconds: 60
       containers:
         - name: {{ .Chart.Name }}
--- a/charts/envoy/templates/deployment.yaml
+++ b/charts/envoy/templates/deployment.yaml
@@ -29,6 +29,9 @@ spec:
     {{- end }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- if .Values.rbac.create }}
+      serviceAccountName: {{ include "envoy.fullname" . }}-envoy
+      {{- end }}
       restartPolicy: Always
       automountServiceAccountToken: false
       terminationGracePeriodSeconds: 60
--- a/charts/envoy/templates/psp-clusterrole.yaml
+++ b/charts/envoy/templates/psp-clusterrole.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.podSecurityPolicy.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "envoy.fullname" . }}-envoy-psp
+  labels:
+    {{- include "envoy.labels" . | nindent 4 }}
+rules:
+- apiGroups: [{{ if .Capabilities.APIVersions.Has "policy/v1beta1" }}'policy'{{ else }}'extensions'{{ end }}]
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - {{ include "envoy.fullname" . }}-envoy-psp
+{{- end }}
--- a/charts/envoy/templates/psp-clusterrolebinding.yaml
+++ b/charts/envoy/templates/psp-clusterrolebinding.yaml
@@ -1,0 +1,16 @@
+{{- if and ( .Values.podSecurityPolicy.enabled ) ( .Values.rbac.create ) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "envoy.fullname" . }}-envoy-psp
+  labels:
+    {{- include "envoy.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "envoy.fullname" . }}-envoy-psp
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "envoy.fullname" . }}-envoy
+    namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/charts/envoy/values.yaml
+++ b/charts/envoy/values.yaml
@@ -69,6 +69,12 @@ service:
       # service.ports.envoy-priv.protocol -- envoy protocol
       protocol: TCP
 
+rbac:
+  # rbac.create -- If true, create and use RBAC resources
+  create: true
+  # rbac.serviceAccountAnnotations -- Annotations for the Service Account
+  serviceAccountAnnotations: {}
+
 podSecurityPolicy:
   # podSecurityPolicy.enabled -- enable pod security policy
   enabled: true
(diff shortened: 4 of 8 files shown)


charts/envoy/values.yaml
```diff
@@ -32,13 +32,22 @@ podSecurityContext: {}
...
-  # readOnlyRootFilesystem: true
-  # runAsNonRoot: true
-  # runAsUser: 1000
+securityContext:
+  # securityContext.capabilities -- Capabilities (specifically, Linux capabilities), are used for permission management in Linux. Some capabilities are enabled by default
+  capabilities:
+    drop:
+      - ALL
+    add:
+      - NET_BIND_SERVICE
```
miseyu, line 41: Because `NET_BIND_SERVICE` needs to be used with envoy

charts/envoy/templates/deployment.yaml
```diff
@@ -26,6 +30,7 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       restartPolicy: Always
+      automountServiceAccountToken: false
```
miseyu, line 33: [error] AutomountServiceAccountTokenTrueAndDefaultSA

charts/envoy/values.yaml
```diff
@@ -32,13 +32,22 @@ podSecurityContext: {}
...
-securityContext: {}
-  # capabilities:
-  #   drop:
-  #   - ALL
-  # readOnlyRootFilesystem: true
-  # runAsNonRoot: true
-  # runAsUser: 1000
+securityContext:
+  # securityContext.capabilities -- Capabilities (specifically, Linux capabilities), are used for permission management in Linux. Some capabilities are enabled by default
+  capabilities:
```
miseyu, line 37: [error] CapabilityOrSecurityContextMissing

charts/envoy/values.yaml
```diff
@@ -32,13 +32,22 @@ podSecurityContext: {}
...
+    drop:
+      - ALL
+    add:
+      - NET_BIND_SERVICE
+  # securityContext.runAsNonRoot -- Containers should be run as a non-root user with the minimum required permissions (principle of least privilege)
+  runAsNonRoot: true
```
miseyu, line 43: [error] RunAsNonRootPSCNilCSCNil

charts/envoy/values.yaml
```diff
@@ -32,13 +32,22 @@ podSecurityContext: {}
...
+  # securityContext.allowPrivilegeEscalation -- AllowPrivilegeEscalation controls whether a process can gain more privileges than its parent process
+  allowPrivilegeEscalation: false
```
miseyu, line 45: [error] AllowPrivilegeEscalationNil

charts/envoy/values.yaml
```diff
@@ -32,13 +32,22 @@ podSecurityContext: {}
...
+
+# podAnnotations -- Pod annotations for the envoy Deployment
+podAnnotations:
+  seccomp.security.alpha.kubernetes.io/pod: 'runtime/default'
```
miseyu, line 49: [error] SeccompAnnotationMissing

charts/envoy/templates/psp-clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: {{ include "envoy.fullname" . }}-envoy-psp
+  labels:
+    {{- include "envoy.labels" . | nindent 4 }}
+rules:
+- apiGroups: [{{ if .Capabilities.APIVersions.Has "policy/v1beta1" }}'policy'{{ else }}'extensions'{{ end }}]
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - {{ include "envoy.fullname" . }}-envoy-psp
```
miseyu, line 13: PodSecurityPolicy Name

charts/envoy/templates/psp-clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.podSecurityPolicy.enabled }}
```
miseyu, line 1: Role for using PodSecurityPolicy

charts/envoy/templates/psp-clusterrolebinding.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if and ( .Values.podSecurityPolicy.enabled ) ( .Values.rbac.create ) }}
```
miseyu, line 1: Role for using PodSecurityPolicy by linking it to the service account used for deployment.

superbrothers (approved):
LGTM! :+1:

superbrothers:
In addition, it would be better to consider running kubeaudit on CI.

miseyu:
@superbrothers 
Thank you very much for your review.:bow:

> In addition, it would be better to consider running kubeaudit on CI.

Integration of `kubeaudit` into CI will be handled separately.

charts/envoy/values.yaml
```diff
@@ -32,13 +32,22 @@ podSecurityContext: {}
...
 
 # securityContext -- Setting security context at the pod applies those settings to all containers in the pod
-securityContext: {}
-  # capabilities:
-  #   drop:
-  #   - ALL
-  # readOnlyRootFilesystem: true
-  # runAsNonRoot: true
-  # runAsUser: 1000
+securityContext:
```
feeblefakie, line 35: Sorry, I'm a bit confused. 
`securityContext` defined here seems not referred whereas `podSecurityContext` above is empty but it seems to be referred by the following.
 Is it OK?
https://github.com/scalar-labs/helm-charts/pull/49/files#diff-f9e366b1cdbdff67956aecad7953e831aed6c1bc6e0b798f3c5fa59845bb5983R31

feeblefakie (commented):
Thank you! Left one question.

charts/envoy/values.yaml
miseyu: @feeblefakie 

Yes, the above configuration is fine.

The corresponding security context for the entire pod is declared as `PodSecurityContext`.
In contrast, the `SecurityContext` is the Security Context of an individual Container.
If there is any overlap, the container's security context will take precedence.

Also, the properties that can be set are different between the two, and the container's security context has more properties that can be set.
In this case, the security context of the container is used.

There is currently only one scalar-envoy container, so we believe that it can meet the requirements.

miseyu:
@feeblefakie 
Please check the answers to your questions. 🙇 
PTAL

charts/envoy/values.yaml
feeblefakie: Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #55, opened 2021-12-08
Add k8s support version for CI

miseyu:
## Summary
- Since EKS and AKS support 1.21, we have added the corresponding k8s version 1.21.
- The reason why it is 1.21.2 is that the latest version of the image of `kindest:node` is `1.21.2
- https://hub.docker.com/r/kindest/node/tags?page=1&name=1.21


feeblefakie (approved):
LGTM! Thank you!

PR #68, opened 2021-12-23
Fix the `values.schema.json` type being `null` 

miseyu:
## Summary
- The type of `values.schema.json` was set to `null` in some places, so this has been fixed.
- `values.yaml` should not be set to `null` as default value
- As for the conditional judgment of `helm`, if it is `null` or `empty character`, it will be `false`, so there is no need to change the logic of template.
- The version that applies `values.schema.json` has not been released yet, so it will not affect existing versions.


feeblefakie (approved):
LGTM! Thank you!

PR #74, opened 2022-01-20
Use scalardl-schema-loader 3.2.1

supl:
This PR tries to release scalar-labs/schema-loading version 2.2.1.
Because version 2.2.0 uses `ghcr.io/scalar-labs/scalardl-schema-loader:3.2.0` which has a bug using a wrong version of scalardb-schema-loader.


supl:
Close due to the wrong release workflow.

PR #76, opened 2022-01-24
Fix threshold for liveness probe

miseyu:
## Summary
- Since the liveness probe may time out depending on the machine status, the threshold of the liveness probe has been changed.


kota2and3kan:
Recently, we supported `startupProbe` in Scalar Helm Chart instead of updating the threshold configuration to address the timeout issue depending on the machine status.
So, I will close this PR. Thanks again for your contribution!

PR #77, opened 2022-01-24
Check for bump up scalardl to v4.1.1

kota2and3kan:
I changed version of scalardl for release new version.
```
$ git checkout -b check-release-v4.1.1 refs/tags/scalardl-4.1.0
```

Could you please confirm the changes?

If there is no problem, I will create prepare-release-v4.1.1 branch in the same way, for release new version. Also, I will release each minor version of scalardl, and release scalardl-audit in the same way.

### DIFF
--- a/charts/scalardl/Chart.yaml
+++ b/charts/scalardl/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: scalardl
 description: Scalar DL is a tamper-evident and scalable distributed database.
 type: application
-version: 4.1.0
-appVersion: 3.3.1
+version: 4.1.1
+appVersion: 3.3.2
 deprecated: false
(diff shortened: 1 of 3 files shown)


miseyu (approved):

kota2and3kan:
Thank you for confirmation!
I will create each `prepare-release-*` branch in the same way.

feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@miseyu In this case, there is no difference, but we should have created a branch from main?

miseyu:
@feeblefakie 
You're right.:bow:

kota2and3kan:
I'm sorry for the mistake.🙇
I will be careful about branch that I create from.

PR #79, opened 2022-01-25
Add scalardl v3.0.1 entry to index.yaml

kota2and3kan:
The entry of scalardl v3.0.1 removed by the following commit. (The following commit is related to scalardl-audit v1.0.1. At this time
we don't find the cause that this commit removes other chart's entry.)
b2a9e78995c8e2d4f99a58859b0a68d305acf84e

I the result of the above behavior, we cannot find chart "scalardl v3.0.1" in the helm search repo command.
So, we need to add entry of "scalardl v3.0.1" into the index.yaml.

Usually, the entry of index.yaml added by chart releaser automatically.
However, to fix the above issue, add entry of scalardl v3.0.1 manually in this commit.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -870,6 +870,37 @@ entries:
     - https://github.com/scalar-labs/helm-charts/releases/download/scalardl-3.1.0/scalardl-3.1.0.tgz
     version: 3.1.0
   - apiVersion: v2
+    appVersion: 3.0.5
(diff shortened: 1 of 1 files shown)


miseyu (approved):
Thank you :bow:

feeblefakie (approved):
LGTM! Thank you!

PR #81, opened 2022-03-09
Fix Scalar DB's Grafana Dashboard

kota2and3kan:
This PR includes the following fixes of Grafana Dashboard of Scalar DB. Please take a look!

## Fix wrong metric name
* Remove `service_` from some metrics name.
* About the following two metrics, rename `service_` to `transaction_`.
    * Abort count
        * Before : `scalardb_stats_distributed_transaction_service_abort_count`
        * After  : `scalardb_stats_distributed_transaction_transaction_abort_count`
    * Abort
        * Before : `scalardb_stats_distributed_transaction_service_abort`
        * After  : `scalardb_stats_distributed_transaction_transaction_abort`

## Change tilet of row
* "Total" row includes requests rate of success and failure. So, it seems that the "Total Requests Rate" is better.
    * "Total" -> "Total Requests Rate"
* It seems that the "Distributed Transaction" is better.
    * "Distributed Storage Transaction Service" -> "Distributed Transaction Service"

## Change title of panel
* Some panel uses the irate() function of PromQL. It returns the *Rate* in the specific period (1 minute). It does not show total *Count*. So, I renamed panel title from `count` to `rate per one minute`.
    * "Total Succeded" -> "Rate of Success Requests per one minute"
    * "Total Failed" -> "Rate of Failure Requests per one minute"
    * Others
        * e.g.) "Get count" -> "Get rate per one minute"
    * (Reference) https://prometheus.io/docs/prometheus/latest/querying/functions/#irate
* Scalar DB exposes the execution time of each gRPC endpoint. However, the title of panel is simple and confusing. So, I add `execution time (percentile)` to each title.
    * e.g.) "Get" -> "Get execution time (percentile)"

## Graph Capture after fix
* [Grafana_Dashboard_Capture.zip](https://github.com/scalar-labs/helm-charts/files/8213084/Grafana_Dashboard_Capture.zip)

### DIFF
--- a/charts/scalardb/files/grafana/scalardb_grafana_dashboard.json
+++ b/charts/scalardb/files/grafana/scalardb_grafana_dashboard.json
@@ -29,7 +29,7 @@
       },
       "id": 13,
       "panels": [],
-      "title": "Total Requests Rate",
+      "title": "Total Requests",
       "type": "row"
     },
     {
@@ -89,7 +89,7 @@
       "timeFrom": null,
       "timeRegions": [],
       "timeShift": null,
-      "title": "Rate of Success Requests per one minute",
+      "title": "Success Requests per one second",
       "tooltip": {
         "shared": true,
         "sort": 0,
@@ -183,7 +183,7 @@
       "timeFrom": null,
       "timeRegions": [],
       "timeShift": null,
-      "title": "Rate of Failure Requests per one minute",
+      "title": "Failure Requests per one second",
       "tooltip": {
         "shared": true,
         "sort": 0,
@@ -311,7 +311,7 @@
           "refId": "G"
         }
       ],
-      "title": "Get rate per one minute",
+      "title": "Get per one second",
       "type": "timeseries"
     },
     {
@@ -511,7 +511,7 @@
           "refId": "G"
         }
       ],
-      "title": "Open scanner rate per one minute",
+      "title": "Open scanner per one second",
       "type": "timeseries"
     },
     {
@@ -711,7 +711,7 @@
           "refId": "G"
         }
       ],
-      "title": "Scan next rate per one minute",
+      "title": "Scan next per one second",
       "type": "timeseries"
     },
     {
@@ -911,7 +911,7 @@
           "refId": "G"
         }
       ],
-      "title": "Mutate rate per one minute",
+      "title": "Mutate per one second",
       "type": "timeseries"
     },
     {
@@ -1125,7 +1125,7 @@
           "refId": "G"
         }
       ],
-      "title": "Create table rate per one minute",
+      "title": "Create table per one second",
       "type": "timeseries"
     },
     {
@@ -1325,7 +1325,7 @@
           "refId": "G"
         }
       ],
-      "title": "Drop table rate per one minute",
+      "title": "Drop table per one second",
       "type": "timeseries"
     },
     {
@@ -1525,7 +1525,7 @@
           "refId": "G"
         }
       ],
-      "title": "Truncate table rate per one minute",
+      "title": "Truncate table per one second",
       "type": "timeseries"
     },
     {
@@ -1724,7 +1724,7 @@
           "refId": "A"
         }
       ],
-      "title": "Get table metadata rate per one minute",
+      "title": "Get table metadata per one second",
       "type": "timeseries"
     },
     {
@@ -1938,7 +1938,7 @@
           "refId": "G"
         }
       ],
-      "title": "Transaction start rate per one minute",
+      "title": "Transaction start per one second",
       "type": "timeseries"
     },
     {
@@ -2138,7 +2138,7 @@
           "refId": "G"
         }
       ],
-      "title": "Transaction get rate per one minute",
+      "title": "Transaction get per one second",
       "type": "timeseries"
     },
     {
@@ -2338,7 +2338,7 @@
           "refId": "G"
         }
       ],
-      "title": "Transaction scan rate per one minute",
+      "title": "Transaction scan per one second",
       "type": "timeseries"
     },
     {
@@ -2538,7 +2538,7 @@
           "refId": "G"
         }
       ],
-      "title": "Transaction mutate rate per one minute",
+      "title": "Transaction mutate per one second",
       "type": "timeseries"
     },
     {
@@ -2738,7 +2738,7 @@
           "refId": "G"
         }
       ],
-      "title": "Transaction commit rate per one minute",
+      "title": "Transaction commit per one second",
       "type": "timeseries"
     },
     {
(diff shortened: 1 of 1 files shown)


brfrn169 (approved):
LGTM! Thank you!

feeblefakie:
@kota2and3kan Thank you!
One question. I understand it uses `irate`, but as far as I can see from the capture of `Total_Requests_Rate`, it doesn't look like the sum of the rates is not equal to 100.
Do you know the reason?

kota2and3kan:
This is noted as a reminder about PromQL used in the scalardb chart.  
In the scalardb chart, Grafana uses like the following query (PromQL).  

* Query
  ```
  sum(irate(scalardb_stats_total_success{pod=~\"$pod\"}[1m])) by (pod)
  ```

### About `[1m]`
`[1m]` means that the `scalardb_stats_total_success` return metrics latest 1 minute data.  
By default of scalardb chart, scrape interval is 15s.  
So, `scalardb_stats_total_success[1m]` basically returns 4 data (t, t-15, t-30, t-45).  
And, `scalardb_stats_total_success[2m]` basically returns 8 data (t, t-15, t-30, ..., t-90, t-105)  

However, irate() uses last two point data (t, t-15) only.  
So, value of `irate(scalardb_stats_total_success[1m])` and `irate(scalardb_stats_total_success[2m])` are same.  

* Reference
    * https://prometheus.io/docs/prometheus/latest/querying/basics/#range-vector-selectors

### About `{pod=~\"$pod\"}`
`{pod=~\"$pod\"}` means that `scalardb_stats_total_success` returns metrics that maches label `pod` = `pod_name`.  
And, $pod is valiable that is feature of Grafana.
In the scalardb chart, $pod defined as each pod name.

* References
    * https://prometheus.io/docs/prometheus/latest/querying/basics/#instant-vector-selectors
    * https://grafana.com/docs/grafana/latest/variables/
    * https://grafana.com/docs/grafana/latest/datasources/prometheus/#templating
    * https://github.com/scalar-labs/helm-charts/blob/main/charts/scalardb/files/grafana/scalardb_grafana_dashboard.json#L5891

### About `sum() ... by (pod)`
`sum() ... by (pod)` means summarize total request by each pod. (It looks like GROUP BY clause of SQL.)
Also, it seems that it set label `pod="pod_name"` each metrics based on the related pod name.
And, use the label to specify the pod in the graph.

kota2and3kan:
@feeblefakie 
Thank you for review!  
And, sorry for confusing. I had a misunderstanding...  

The irate() calculates **"increase rate per second"** by using last two point data.  
So, the sum of values (rate) will not necessarily be 100.  

## Behavior of irate() of Prometheus
For example, it shows the following result.  
(Prometheus retrieves metrics every 15 second by default of scalardb chart)  

* Example data
  ```
                                    t1                t2 t3       t4
                                    |                 |  |        |
  time  : 0s        15s       30s   |   45s       60s |  |  75s   |   90s
          +---------+---------+---------+---------+---------+---------+
  count : 0         20        50        60        210       270       300
  ```
  Note: count = total count of request (monotonically increasing value)

* Result
    * irate() of t1 -> (50 - 20) / (30 - 15) = 2
    * irate() of t2 -> (210 - 60) / (60 - 45) = 10
    * irate() of t3 -> (210 - 60) / (60 - 45) = 10
        * Because the `last two point` of t2 and t3 is same, the result same too.
    * irate() of t4 -> (270 - 210) / (75 - 60) = 4

This result will be ploted into the graph of Grafana.  
In other words, it shows "0", if there is no request in the specific period.  

* References (irate() calculates values here)
    * https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L149
    * https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L158

## Behavior of Grafana of scalardb chart (with kube-prometheus-stack)
In the scalardb chart with kube-prometheus-stack, the following values (X/Y axis) will be ploted into graph of Grafana.

* Example of scalardb chart with kube-prometheus-stack
  ```
          Prometheus (interval = 15s by default of scalardb chart)
  time  : 0s    15s   30s   45s   60s   90s   105s  120s  135s  150s  165s
          +-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+
  count : 0     20    50    60    210   240   300   330   330   350   410
            |           |           |           |           |           |
  value     |           |           |           |           |           |
  X axis    t0          t1          t2          t3          t4          t5
  Y axis    0           2           10          4           0           4
          Grafana (interval = 30s by default of kube-prometheus-stack)
  ```
    * t0 : (0 - 0) / 15 = 0
    * t1 : (50 - 20) / 15 = 2
    * t2 : (210 - 60) / 15  = 10
    * t3 : (300 - 240) / 15 = 4
    * t4 : (330 - 330) / 15 = 0
    * t5 : (410 - 350) / 15 = 4

This graph does not represent a perfect **increase rate per second**.
Because, for example, the total requests increased 30 between 105s and 135s, but Grafana shows "0" at t4.
It shows 0, in spite of the total requests increased between t3 and t4.

This behavior caused by different scrape interval between Prometheus and Grafana.
However, scrape interval of Grafana is set by kube-prometheus-stack.
In other words, scrape interval of Grafana depends on each user environment, and we cannot control it completely.
(Prometheus's scrape interval set in the scalardb helm chart.)

## About titles
#### About title of row
Now, I have a feeling that "Total Request Rate" that you pointed out is not appropriate...  
I think "Request Increase Rate" is better than "Total Requests Rate".  

#### About title of panel
Also, I think "XXX increase rate" is better than "XXX rate per one minute" for each panel title.  
(Each panel also uses irate().)  

* e.g.)
    * `Rate of Success Requests per one minute` -> `Success Request increase rate`
    * `Rate of Failure Requests per one minute` -> `Failure Request increase rate`
    * `Get rate per one minute` -> `Get increase rate`
    * `Mutate rate per one minute` -> `Mutate increase rate`
    * `Transaction start rate per one minute` -> `Transaction start increase rate`

According to the specification of irate() function, strictly, `XXX increase rate per second` is correct.  
However, as I mentioned above, Grafana does not shows perfect `increase rate per second`, and we cannot control it completely.  
It is a rough indication of `increase rate per second`.  
If we set `per second`, I think that it might cause misunderstandings for uses.  

So, I think that `XXX increase rate` is better.  
What do you think about these titles?

kota2and3kan:
@feeblefakie 
I fixed the panel titles based on the discussion.

* Remove "rate", becouse the graph shows (PromQL returns) Requests Per Second.
* Change the time period from "minute" to "second", because irate() returns "per one second" value.

The graph I fixed is following.
<details>
<summary>Fixed Graph Capture</summary>

Grafana_Dashborad_All
</details>

Please take a look!

feeblefakie (approved):
LGTM! Thank you!

PR #85, opened 2022-04-19
Add Getting Started Guide with Helm Chart (Scalar DL)

kota2and3kan:
This PR add the getting started guide with helm charts of Scalar DL.

There are two documents.
* Getting Started with Helm Charts (Scalar DL Ledger)
    * Create Ledger environment on the minikube.
* Getting Started with Helm Charts (Scalar DL Auditor)
    * Add the Auditor environment on the same minikube as Ledger.
    * The minikube is same, but I use two Cassandra (for Ledger and Auditor).

I created this document with priority on reproducibility in user's environment.  
So, I describe concrete commands in all steps.  
Also, to avoid issues depending on each environment as much as possible, this guide runs all component (include the client that runs Scalar DL sample contract) as docker container.  

Also, there are some redundancy in the `Run Scalar DL sample contract` section between two document.  
However, there are some differences. For example, property files, regstering cert files, and result of Request (Contract Execution/Validation).
So, I describe details of all steps in that section in both document.  

Please take a look!

PR #86, opened 2022-04-20
Add Getting Started Guide with Helm Chart (Monitoring)

kota2and3kan:
This PR add the getting started guide with helm charts of the monitoring.

There are two files.

* docs/conf/scalar-prometheus-custom-values.yaml
    * This is a minimum custom value file of `kube-prometheus-stack` for monitoring Scalar products.
    * I have created this file based on the following file.
        * https://github.com/superbrothers/helm-charts/blob/monitoring/charts/monitoring/values.yaml
* docs/getting-started-monitoring.md
    * This guide deploys Prometheus Operator (kube-prometheus-stack) on the minikube for testing.
    * Also, this guide describes the setting that we need to enable monitoring of each Scalar product.

Also, there are some redundancy between this guide and `Getting Started with Helm Charts (Scalar DB/DL)`.   
I will separate the redundancy point in the another PR after each guide are merged, because we need to keep consistency between each document.  

Please take a look!

### DIFF
--- a/docs/getting-started-monitoring.md
+++ b/docs/getting-started-monitoring.md
@@ -100,6 +100,11 @@ First, you need to install the following tools used in this guide.
      
      ...
      ```
+   * Note:
+       * If you want to customize the Prometheus Operator deployment using Helm Charts, you need to set the following configuration for monitoring Scalar products.
+           * The `serviceMonitorSelectorNilUsesHelmValues` and `ruleSelectorNilUsesHelmValues` must be set to `false` (`true` by default) to make Prometheus Operator detects `ServiceMonitor` and `PrometheusRule` of Scalar products.
+
+
 
 ## Step 4. Deploy `kube-prometheus-stack`
 
--- a/docs/getting-started-monitoring.md
+++ b/docs/getting-started-monitoring.md
@@ -1,0 +1,294 @@
+# Getting Started with Helm Charts (Monitoring using Prometheus Operator)
+
+This document explains how to get started with monitoring Scalar products on Kubernetes using Prometheus Operator (kube-prometheus-stack). Here, we assume that you already have a Mac or Linux environment for testing.  
+
+## Environment
+
+In this guide, we will create the following environment in your local by using Docker and minikube.  
+
+```
++--------------------------------------------------------------------------------------------------+
+| +------------------------------------------------------+                     +-----------------+ |
+| |                kube-prometheus-stack                 |                     | Scalar Products | |
+| |                                                      |                     |                 | |
+| | +--------------+  +--------------+  +--------------+ | ---(Monitoring)---> |  +-----------+  | |
+| | |  Prometheus  |  | Alertmanager |  |   Grafana    | |                     |  | Scalar DB |  | |
+| | +-------+------+  +------+-------+  +------+-------+ |                     |  +-----------+  | |
+| |         |                |                 |         |                     |  +-----------+  | |
+| |         +----------------+-----------------+         |                     |  | Scalar DL |  | |
+| |                          |                           |                     |  +-----------+  | |
+| +--------------------------+---------------------------+                     +-----------------+ |
+|                            |                                                                     |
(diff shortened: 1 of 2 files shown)


docs/getting-started-monitoring.md
```diff
@@ -0,0 +1,294 @@
...
+                             | <- expose to localhost (127.0.0.1)
+                             |
+              (Access Dashboard through HTTP)
+                             |
+                        +----+----+
+                        | Browser |
+                        +---------+
+```
+
+## Step 1. Install tools
```
kota2and3kan, line 33: This section is a redundancy of the `Getting Started with Helm Charts (Scalar DB/DL)`.  
So, I will separate this section in the another PR after each guide are merged.

kota2and3kan:
@scalar-boney  
As well as #85, we have already confirmed that the tools works on Mac.  
So, we don't need to test this document on Mac.  

Also, redundancy points are remains in this guide (`Install tools` and `Start minikube with docker driver`).  
However, I will separate these sections in the another PR after each guide are merged.  

Could you please review this guide in the perspective of documentation (e.g. some errors in the sentence)?

docs/getting-started-monitoring.md
```diff
@@ -0,0 +1,294 @@
+# Getting Started with Helm Charts (Monitoring using Prometheus Operator)
+
+This document explains how to get started with monitoring Scalar products on Kubernetes using Prometheus Operator (kube-prometheus-stack). Here, we assume that you already have a Mac or Linux environment for testing.
```
scalar-boney, line 3:
```suggestion
This document explains how to get started with Scalar products monitoring on Kubernetes using Prometheus Operator (kube-prometheus-stack). Here, we assume that you already have a Mac or Linux environment for testing.  
```

scalar-boney (approved):
LGTM!

Left one minor suggestion

docs/getting-started-monitoring.md
kota2and3kan: Thank you for your review!
I applied the suggestion.

kota2and3kan:
I reflected the changes based on the comments on the other guide in 3ba64b6c5133314eaadfce6e2897760f8c68f252.

kota2and3kan:
I removed the prompt from each code snippet (command) in 31073db1ebc7bac4028694df2cf1cc5d3bf06c42.
After this fixing, users can copy command from button in the top of right in each code snippet, and paste directly.

Also, I separated the execution result in the code snippet, because we don't need copy it.

kota2and3kan:
@feeblefakie  
Thank you for your review!

> I don't remember what it is but I think we have some configuration to enable the Prometheus operator to work. Is it OK without it?

By default, the value of `serviceMonitorSelectorNilUsesHelmValues` and `ruleSelectorNilUsesHelmValues` are `true`.
So, the Prometheus Operator doesn't detect `ServiceMonitor` and `PrometheusRule` of scalar products by default.

* Example (Scalar DB)
  * [ServiceMonitor](https://github.com/scalar-labs/helm-charts/blob/scalardb-2.2.4/charts/scalardb/templates/scalardb/servicemonitor.yaml)
  * [PrometheusRule](https://github.com/scalar-labs/helm-charts/blob/scalardb-2.2.4/charts/scalardb/templates/scalardb/prometheusrules.yaml)

We need to set `false` to the above configuration to make Prometheus Operator detects `ServiceMonitor` and `PrometheusRule` of scalar products.
If you use [sample custom value file](https://github.com/scalar-labs/helm-charts/blob/ed74d88cd220fd279c36663d52e13902d76049a1/docs/conf/scalar-prometheus-custom-values.yaml), the above parameters are set `false`.
However, if you want to customize the Prometheus Operator (or you already have a Prometheus Operator deployment using Helm Charts) using your (user's) original custom value file, you need to set those configurations to `false`.

So, I added the note about it in ed74d88cd220fd279c36663d52e13902d76049a1.

PR #92, opened 2022-05-23
Update version of each chart to latest

kota2and3kan:
This PR update version of each chart to keep main branch latest.
We need to update version in main branch when we release most recent version of Helm Charts to keep main branch latest, but I forgot it.
Sorry for overlooking this issue...

Also, we don't need to do any release operation because all latest charts are released correctly as follows.
```console
$ helm search repo scalar-labs
NAME                            CHART VERSION   APP VERSION     DESCRIPTION
scalar-labs/envoy               2.0.1           1.2.0           Envoy Proxy for Scalar applications
scalar-labs/scalardb            2.2.4           3.5.2           Scalar DB server
scalar-labs/scalardl            4.2.1           3.4.0           Scalar DL is a tamper-evident and scalable dist...
scalar-labs/scalardl-audit      2.2.1           3.4.0           Scalar DL is a tamper-evident and scalable dist...
scalar-labs/schema-loading      2.5.1           3.4.0           A schema loading tool for Scalar DL.
```

The purpose of this PR is only keeping main branch latest.
Please take a look.
(I will update the docs about release flow in the another PR.)

PR #93, opened 2022-05-23
Update release flow document

kota2and3kan:
This PR update the release flow document to keep main branch latest.
This PR is related to the another PR #92.

Please take a look!

### DIFF
--- a/docs/ReleaseFlow.md
+++ b/docs/ReleaseFlow.md
@@ -39,3 +39,114 @@ $ git push origin prepare-release-v3.0.2
       * If you release a new chart v2.1.6 (Scalar DB v3.4.x), you don't need to update the version of the main branch.
       * If you release a new chart v2.2.5 (Scalar DB v3.5.x), you need to update the version of the main branch.
       * Also, if you release a new minor (v2.3.0) or major (v3.0.0) chart (Scalar DB v3.x or v4.x), you need to update the version of the main branch.
+    * What you need to update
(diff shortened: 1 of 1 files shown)


docs/ReleaseFlow.md
```diff
@@ -22,3 +22,20 @@ $ git push origin prepare-release-v3.0.2
 ```
 
 * In this case, the `Github Actions` for the release will be executed upon `Git push`.
+
+* If you release the most recent version of charts, you need to update the version of the main branch as well.
```
feeblefakie, line 26: Can you clarify what needs to be updated?
Also just wondering if it can be done automatically?

kota2and3kan: > Can you clarify what needs to be updated?

I added the descriptions about what we need to update in 3059edffc83db53459990a30dc77d14f67a9838b.

However, I think these steps are almost **release steps** (they are very similar to release steps).
So, I think that it is better to update entire docs to explain release steps in more detail.

> Also just wondering if it can be done automatically?

I think there is no way to automatically update now.

As manually ways, I think there are two options as follows.
* Create PR to keep main branch latest after releasing latest charts.
* Trigger release workflow by merging into the main, only when we wil release latest version charts.

For update the docs that as mentioned above, I would like to discuss about release flow.
Can I talk to you about this PR in the next week?

kota2and3kan:
According to the result of discussion, I close this PR.
And, I will test a new release flow.

PR #94, opened 2022-05-23
Configure Grafana to allow the embedding feature

supl:
This PR adds a configuration to the sample custom values of the "Getting Started Monitoring" guide to allow Grafana to be embedded in the other websites (mainly, for Scalar Manager by far).

### DIFF
--- a/docs/conf/scalar-prometheus-custom-values.yaml
+++ b/docs/conf/scalar-prometheus-custom-values.yaml
@@ -26,6 +26,13 @@ grafana:
     # requests:
     #   memory: 400Mi
 
+  # -- Grafana's primary configuration
+  grafana.ini:
+    security:
+      # -- allow Grafana to be embedded (not set the X-Frame-Options header)
+      # -- https://grafana.com/docs/grafana/latest/administration/configuration/#allow_embedding
+      allow_embedding: true
+
 kubeApiServer:
   # -- Scraping kube-apiserver is disabled
   enabled: false
--- a/docs/conf/scalar-prometheus-custom-values.yaml
+++ b/docs/conf/scalar-prometheus-custom-values.yaml
@@ -30,8 +30,9 @@ grafana:
   grafana.ini:
     security:
       # -- allow Grafana to be embedded (not set the X-Frame-Options header)
+      # -- If you use Scalar Manager, you need to set allow_embedding to true.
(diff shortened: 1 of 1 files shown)


docs/conf/scalar-prometheus-custom-values.yaml
```diff
@@ -26,6 +26,13 @@ grafana:
     # requests:
     #   memory: 400Mi
 
+  # -- Grafana's primary configuration
+  grafana.ini:
```
supl, line 30: This part can be found in
https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L577

docs/conf/scalar-prometheus-custom-values.yaml
```diff
@@ -26,6 +26,13 @@ grafana:
...
+    security:
+      # -- allow Grafana to be embedded (not set the X-Frame-Options header)
+      # -- https://grafana.com/docs/grafana/latest/administration/configuration/#allow_embedding
+      allow_embedding: true
```
kota2and3kan, line 34:
```suggestion
      # -- allow Grafana to be embedded (not set the X-Frame-Options header)
      # -- If you use Scalar Manager, you need to set allow_embedding to true.
      # -- https://grafana.com/docs/grafana/latest/administration/configuration/#allow_embedding
      allow_embedding: false
```
According to the following doc, this configuration mitigates the risk of Clickjacking. 
https://grafana.com/docs/grafana/latest/administration/configuration/#allow_embedding

Also, I think we use this sample custom values file in several cases other than Scalar Manager. 
And, we don't need to set `true` when we don't use Scalar Manager.

So, I think it is better to set `false` by default and set `true` when we use Scalar Manager only.
I think we can add a comment that we need to set `true` when we use Scalar Manager in the sample custom values file.

What do you think?

kota2and3kan (commented):
Thank you for the update!
I left a comment. Please take a look.

docs/conf/scalar-prometheus-custom-values.yaml
supl: @kota2and3kan 
Thanks for your feedback.
I agree with your concerns.
It should be set with the default value `false` in the sample custom values.
I will set it `true` as long as we really need it.

feeblefakie (approved):
LGTM! Thank you!

kota2and3kan (approved):
LGTM! Thank you!

PR #95, opened 2022-05-23
[WIP] Use secret resource for usename and password in Scalar DB chart

kota2and3kan:
* Note: 
    * This is a PR for discussion.
    * If we choose this way, we need to apply this update after releasing new version Scalar DB that includes the following PR.
      https://github.com/scalar-labs/scalardb/pull/590

---

This PR update Scalar DB Server Helm Charts to pass the username and password via Secret resource.
In this PR, we use environment variable (Secret) for username/password only.
Another values are passed as ConfigMap.

I would like to discuss which of the following two is better.
* Use secret (environment variable) for username/password only (This PR).
* Use environment variable for all configurations as well as [Scalar DL](https://github.com/scalar-labs/helm-charts/blob/scalardl-4.2.1/charts/scalardl/templates/ledger/deployment.yaml#L54-L87).

### DIFF
--- a/charts/scalardb/templates/scalardb/configmap.yaml
+++ b/charts/scalardb/templates/scalardb/configmap.yaml
@@ -8,8 +8,8 @@ data:
   database.properties: |-
     scalar.db.contact_points={{.Values.scalardb.storageConfiguration.contactPoints}}
     scalar.db.contact_port={{.Values.scalardb.storageConfiguration.contactPort}}
-    scalar.db.username={{.Values.scalardb.storageConfiguration.username}}
-    scalar.db.password={{.Values.scalardb.storageConfiguration.password}}
+    scalar.db.username={{ "{{ default .Env.SCALAR_DB_USERNAME \"\" }}" }}
+    scalar.db.password={{ "{{ default .Env.SCALAR_DB_PASSWORD \"\" }}" }}
     scalar.db.storage={{.Values.scalardb.storageConfiguration.storage}}
     scalar.db.server.port=50051
 ---
--- a/charts/scalardb/templates/scalardb/deployment.yaml
+++ b/charts/scalardb/templates/scalardb/deployment.yaml
@@ -49,6 +49,24 @@ spec:
           env:
           - name: JAVA_OPTS
             value: "-Dlog4j.logLevel={{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
+          - name: SCALAR_DB_USERNAME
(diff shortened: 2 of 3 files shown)


charts/scalardb/templates/scalardb/configmap.yaml
```diff
@@ -8,8 +8,8 @@ data:
   database.properties: |-
     scalar.db.contact_points={{.Values.scalardb.storageConfiguration.contactPoints}}
     scalar.db.contact_port={{.Values.scalardb.storageConfiguration.contactPort}}
-    scalar.db.username={{.Values.scalardb.storageConfiguration.username}}
-    scalar.db.password={{.Values.scalardb.storageConfiguration.password}}
+    scalar.db.username={{ "{{ default .Env.SCALAR_DB_USERNAME \"\" }}" }}
+    scalar.db.password={{ "{{ default .Env.SCALAR_DB_PASSWORD \"\" }}" }}
```
kota2and3kan, line 12: These values will be evaluated as follows by Helm.
```
```

And, the above configuration (strings) will be mounted to container as `database.properties.tmpl`.
After that, the `dockerize` evaluates these values (replace to actual username and password) based on the environment variables (SCALAR_DB_USERNAME/SCALAR_DB_PASSWORD) as follows.

```
$ cat << EOF > test.properties 
EOF
```
```
$ ./dockerize -template ./test.properties
scalar.db.username=
scalar.db.password=
```
```
$ export SCALAR_DB_USERNAME=test-user
$ export SCALAR_DB_PASSWORD=test-password
$ ./dockerize -template ./test.properties
scalar.db.username=test-user
scalar.db.password=test-password
```

Also, the above environment variables will be passed as Secret resource.

charts/scalardb/templates/scalardb/deployment.yaml
```diff
@@ -63,8 +81,8 @@ spec:
             timeoutSeconds: 1
           volumeMounts:
             - name: scalardb-config-file-volume
-              mountPath: /scalardb/server/database.properties
-              subPath: database.properties
+              mountPath: /scalardb/server/database.properties.tmpl
+              subPath: database.properties.tmpl
```
kota2and3kan, line 85: For `dockerize` command runs in the Scalar DB Dockerfile, we need to mount properties file as `database.properties.tmpl`.

kota2and3kan:
As a result of discussion, we decided to implement this feature in the another way.
So, I close this PR.

PR #96, opened 2022-05-23
Support customized database.properties in Scalar DB

kota2and3kan:
This PR add a new feature that accepting user customized database.properties file in the Scalar DB Helm Charts.
By default, it is disabled.

Please take a look!

Note that we need to apply this update after releasing new version Scalar DB that includes the following PR. In other words, if you test this feature, you need to use the Scalar DB Server 4.0.0-SNAPSHOT at this time.
* https://github.com/scalar-labs/scalardb/pull/590

---

## How to use your customized database.properties file

You can use your customized database.properties by the following steps.

1. Set `scalardb.useCustomizedConfiguration.enabled=true` in your custom value file.
1. Create a `ConfigMap` resource using your database.properties file as follows.
   ```console
   kubectl create configmap scalardb-customized-config --from-file=database.properties.tmpl=database.properties
   ```
1. Deploy Scalar DB using Helm Charts.

---

## How to pass the credentials as a environment variables using `Secret` resource

Also, you can use a `Secret` resource to pass the credentials as a environment variables as follows.

1. Set `scalardb.useCustomizedConfiguration.useSecret=true` in your custom value file.
1. Create the `ConfigMap` includes environment variable name as Go template syntax.
   ```
   scalar.db.storage=multi-storage
   scalar.db.multi_storage.storages=storage0,storage1
   
   scalar.db.multi_storage.storages.storage0.storage=jdbc
   scalar.db.multi_storage.storages.storage0.contact_points=jdbc:postgresql://postgresql-scalardb-1:5432/postgres
   scalar.db.multi_storage.storages.storage0.username={{ default .Env.SCALAR_DB_POSTGRES_USERNAME_A "" }}
   scalar.db.multi_storage.storages.storage0.password={{ default .Env.SCALAR_DB_POSTGRES_PASSWORD_A "" }}
   
   scalar.db.multi_storage.storages.storage1.storage=jdbc
   scalar.db.multi_storage.storages.storage1.contact_points=jdbc:postgresql://postgresql-scalardb-2:5432/postgres
   scalar.db.multi_storage.storages.storage1.username={{ default .Env.SCALAR_DB_POSTGRES_USERNAME_B "" }}
   scalar.db.multi_storage.storages.storage1.password={{ default .Env.SCALAR_DB_POSTGRES_PASSWORD_B "" }}
   
   scalar.db.multi_storage.namespace_mapping=s0:storage0,s1:storage1
   scalar.db.multi_storage.default_storage=storage0
   ```
1. Create the `Secret` includes credentials.
   ```console
   kubectl create secret generic scalardb-customized-secret \
     --from-literal=db-username-a=postgres \
     --from-literal=db-password-a=aaaaaaaa \
     --from-literal=db-username-b=postgres \
     --from-literal=db-password-b=bbbbbbbb
   ```
1. Specify the **environment variable name** and **key name** of `Secret` in the custom values file.
   ```yaml
   scalardb:
     useCustomizedConfiguration:
       enabled: true
       useSecret: true
       secretKeys:
         - environmentVariableName: SCALAR_DB_POSTGRES_USERNAME_A
           secretKeyName: db-username-a
         - environmentVariableName: SCALAR_DB_POSTGRES_PASSWORD_A
           secretKeyName: db-password-a
         - environmentVariableName: SCALAR_DB_POSTGRES_USERNAME_B
           secretKeyName: db-username-b
         - environmentVariableName: SCALAR_DB_POSTGRES_PASSWORD_B
           secretKeyName: db-password-b
   ```

As a result, the credentials are set in the database.properties file in each container as follows.
```
scalar.db.storage=multi-storage
scalar.db.multi_storage.storages=storage0,storage1

scalar.db.multi_storage.storages.storage0.storage=jdbc
scalar.db.multi_storage.storages.storage0.contact_points=jdbc:postgresql://postgresql-scalardb-1:5432/postgres
scalar.db.multi_storage.storages.storage0.username=postgres
scalar.db.multi_storage.storages.storage0.password=aaaaaaaa

scalar.db.multi_storage.storages.storage1.storage=jdbc
scalar.db.multi_storage.storages.storage1.contact_points=jdbc:postgresql://postgresql-scalardb-2:5432/postgres
scalar.db.multi_storage.storages.storage1.username=postgres
scalar.db.multi_storage.storages.storage1.password=bbbbbbbb

scalar.db.multi_storage.namespace_mapping=s0:storage0,s1:storage1
scalar.db.multi_storage.default_storage=storage0
```


kota2and3kan:
Note: 
CI (Chart Testing and Kubeval / install-chart) failed, but the error is timed out as follows.
```
>>> helm install scalardb-5rj2ovdxks charts/scalardb --namespace default --wait --timeout 300s
Error: INSTALLATION FAILED: timed out waiting for the condition
```
So, I re-ran it manually and it was successful.

feeblefakie (commented):
The feature-wise, it looks great! Thank you!
Left comments on variable naming.

charts/scalardb/values.yaml
```diff
@@ -123,6 +123,22 @@ scalardb:
     # -- The log level of Scalar DB
     dbLogLevel: INFO
 
+  useCustomizedConfiguration:
```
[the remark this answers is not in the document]
kota2and3kan: Thank you for your review!

> First, using `custom` is probably more appropriate in English.

I see. I will use `custom` as prefix of this parameter. Thank you.

> Second, I think `custom configuration` might be too general, so I think it's better to say it's a custom configuration for `strorageConfiguration`.

Indeed, it is too general. So, I agree with the another name is better.
However, I want to discuss about the word `Storage` is appropriate or not.

After this update, basically we are free to change all settings of Scalar DB Server.
For example, we can change the properties like `scalar.db.server.***` and `scalar.db.consensus_commit.***`.
I feel these are not `storage` configuration.
Also, if some parameters that are not related to `storage` will be implemented in the future, we can change them too.

So, I think `customScalardbConfiguration` is better.

> Lastly, `useXXX` might be misleading since it makes me feel it is a single boolean config.

I agree with you. The `useXXX` looks like boolean config. Thank you.

---

Also, I want to discuss about the name of parameter in the Ledger and Auditor charts.

Before, I thought the same name between several charts is more simple and understandable.
However, I agree with it (`custom configuration`) is too general.
And, according to the same reason (we are free to change all settings), I think it is better to not use the word `storage`.

So, I think the following names are better.

* Scalar DB Server -> `customScalardbConfiguration`
* Ledger -> `customLedgerConfiguration`
* Auditor -> `customAuditorConfiguration`

What do you think?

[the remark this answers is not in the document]
ymorimo: Since there is `storageConfiguration` above in the same file, the name `customStorageConfiguration` sounds like a variant of that object to me.
In this chart, the point of the config is to specify a ConfigMap as a template that will override the properties file path `/scalardb/server/database.properties.tmpl` (which is fixed by the Dockerfile) of the scalardb-server container.
I think a name that could represent this point would be nice, but I haven't come up with one.

Also, we might need to consider that when this mechanism is applied to other charts, the name can be reused or derived naturally.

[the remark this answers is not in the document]
ymorimo: Sorry, I posted my comment without reloading @kota2and3kan san's reply.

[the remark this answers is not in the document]
ymorimo: IMHO, it sounds somewhat inaccurate or non-perfect to me to call this `custom`. Specifying the config values with environment variables in the old way is also "custom", isn't it?

If `useCustomizedConfiguration` is enabled, `storageConfiguration` is completely ignored, right? They are exclusive of each other? If that is true, I wonder if it is a good idea to keep two different ways of configuration. (Not sure if this is a positive suggestion, though.)

charts/scalardb/values.yaml
```diff
@@ -123,6 +123,22 @@ scalardb:
...
+    # -- Use user customized database.properties. You need to create ConfigMap includes database.properties.
+    enabled: false
+    # -- ConfigMap name that includes database.properties.
+    configMapName: "scalardb-customized-config"
+    # -- Use Secret to pass the credentials as environment variable.
+    useSecret: false
```
superbrothers, line 132: Is this setting necessary? Do you expect to pass credentials using a resource other than Secret?

charts/scalardb/values.yaml
```diff
@@ -123,6 +123,22 @@ scalardb:
...
+    # -- Secret name that includes credentials.
+    secretName: "scalardb-customized-secret"
+    # -- Array of hash that includes environment variable name and secret key name.
+    secretKeys:
+      - environmentVariableName: SCALAR_DB_USERNAME
+        secretKeyName: db-username
+      - environmentVariableName: SCALAR_DB_PASSWORD
+        secretKeyName: db-password
```
superbrothers, line 140: You can consider using envFrom. In this case, the key name of the Secret resource becomes the environment variable name, and this setting becomes unnecessary.

- https://kubernetes.io/docs/concepts/configuration/secret/#use-case-as-container-environment-variables

I think there is no need to separate Secret key names from environment variable names.

charts/scalardb/values.yaml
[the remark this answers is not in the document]
superbrothers: I see no problem with "Custom" if it provides a default method. If the current default method is to be deprecated in the future, then the name "Custom" becomes a problem.

Whether "storage" should be added depends on whether the configuration file contains only storage. If the configuration file also includes settings other than "storage", then it must be named generically.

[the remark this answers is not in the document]
superbrothers: Another option is to stop using "Custom" by providing default data for the configuration file.

```yaml
scalardb:
  replicaCount: 3
  storageConfiguration:
    contactPoints: cassandra
  configurationData: |
    # Comma separated contact points. For DynamoDB, the region is specified by this parameter.
    scalar.db.contact_points={{ default .Env.SCALAR_DB_CONTACT_POINTS "" }}

    # Port number for all the contact points. Default port number for each database is used if empty.
    scalar.db.contact_port={{ default .Env.SCALAR_DB_CONTACT_PORT "" }}

    # Credential information to access the database. For Cosmos DB, username isn't used. For DynamoDB, AWS_ACCESS_KEY_ID is specified by the username and AWS_ACCESS_SECRET_KEY is specified by the password.
```

For example, the above. If users want to customize the configuration file, they can simply edit the default data. This way there is no need for a special field called "Custom".

---

There is another advantage to having the configuration file in values.yaml. If users create their own ConfigMap and write the ConfigMap name in values.yaml, then when users update the configuration file data contained in the ConfigMap, the Pods are not re-created and the configuration file updates are not reflected. The user must explicitly recreate the Pods for the configuration file updates to take effect.

If we choose to write the configuration file data in values.yaml, we can automatically trigger the re-creation of Pods because we will be aware that the data has been updated. There is no need for users to explicitly recreate the Pods.

- https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

charts/scalardb/values.yaml
kota2and3kan: > Do you expect to pass credentials using a resource other than Secret?

No, I don't expect using other than Secret.
I use this configuration for the if clause only in the deployment.yaml.
So, I removed this setting.

And, I update the condition of if clause to decide to load environment variable from Secret (use `envFrom`) or not.
If the `scalardb.useCustomizedConfiguration.secretName` provided, we use `envFrom` in the deployment.yaml.

charts/scalardb/values.yaml
kota2and3kan: I didn't know the `envFrom`. Thank you!

I updated the deployment.yaml to use `envFrom`.
Also, I removed this unnecessary settings from values.yaml.

kota2and3kan:
@feeblefakie @ymorimo @superbrothers  
I updated this PR based on the review and discussion.

* Apply the following way to use customized `database.properties`.
  https://github.com/scalar-labs/helm-charts/pull/96#discussion_r889630203
* Use `envFrom` to pass the environment variables from `Secret`.
* Remove unnecessary values from `values.yaml`.

Please take a look again!

---

### For your information

You can use your database.properties as follows.

1. Prepare a custom values file (includes database.properties).
   ```yaml
   scalardb:
   
     databaseProperties: |
       # The storage is "multi-storage"
       scalar.db.storage=multi-storage
   
       # Define storage names, comma-separated format. In this case, "cassandra" and "mysql"
       scalar.db.multi_storage.storages=storage0,storage1
   
       # s1 (coordinator.state)
       scalar.db.multi_storage.storages.storage0.storage=jdbc
       scalar.db.multi_storage.storages.storage0.contact_points=jdbc:postgresql://postgresql-scalardb-1:5432/postgres
       scalar.db.multi_storage.storages.storage0.username={{ default .Env.SCALAR_DB_POSTGRES_USERNAME_A "" }}
       scalar.db.multi_storage.storages.storage0.password={{ default .Env.SCALAR_DB_POSTGRES_PASSWORD_A "" }}
   
       # s2 (emoney.account)
       scalar.db.multi_storage.storages.storage1.storage=jdbc
       scalar.db.multi_storage.storages.storage1.contact_points=jdbc:postgresql://postgresql-scalardb-2:5432/postgres
       scalar.db.multi_storage.storages.storage1.username={{ default .Env.SCALAR_DB_POSTGRES_USERNAME_B "" }}
       scalar.db.multi_storage.storages.storage1.password={{ default .Env.SCALAR_DB_POSTGRES_PASSWORD_B "" }}
   
       # Define namespace mappings from a namespace name to a storage. The format is "<namespace name>:<storage name>,..."
       scalar.db.multi_storage.namespace_mapping=s0:storage0,s1:storage1
   
       # Define the default storage that’s used if a specified table doesn’t have any table mapping
       scalar.db.multi_storage.default_storage=storage0
   
     secretName: "scalardb-customized-secret"
   ```
1. Create `Secret` resource includes credentials if you need.
   ```console
   kubectl create secret generic scalardb-customized-secret \
     --from-literal=SCALAR_DB_POSTGRES_USERNAME_A=postgres \
     --from-literal=SCALAR_DB_POSTGRES_PASSWORD_A=aaaaaaaa \
     --from-literal=SCALAR_DB_POSTGRES_USERNAME_B=postgres \
     --from-literal=SCALAR_DB_POSTGRES_PASSWORD_B=bbbbbbbb
   ```
1. Deploy Scalar DB Server.
   ```console
   helm install scalardb ~/github.com/scalar-labs/helm-charts/charts/scalardb/ -f ./scalardb-server.yaml
   ```

As a result, you can see your customized database.properties in the Pod as follows.
```console
$ kubectl exec -t scalardb-dd55c6b69-d4vwx -- cat database.properties
# The storage is "multi-storage"
scalar.db.storage=multi-storage

# Define storage names, comma-separated format. In this case, "cassandra" and "mysql"
scalar.db.multi_storage.storages=storage0,storage1

# s1 (coordinator.state)
scalar.db.multi_storage.storages.storage0.storage=jdbc
scalar.db.multi_storage.storages.storage0.contact_points=jdbc:postgresql://postgresql-scalardb-1:5432/postgres
scalar.db.multi_storage.storages.storage0.username=postgres
scalar.db.multi_storage.storages.storage0.password=aaaaaaaa

# s2 (emoney.account)
scalar.db.multi_storage.storages.storage1.storage=jdbc
scalar.db.multi_storage.storages.storage1.contact_points=jdbc:postgresql://postgresql-scalardb-2:5432/postgres
scalar.db.multi_storage.storages.storage1.username=postgres
scalar.db.multi_storage.storages.storage1.password=bbbbbbbb

# Define namespace mappings from a namespace name to a storage. The format is "<namespace name>:<storage name>,..."
scalar.db.multi_storage.namespace_mapping=s0:storage0,s1:storage1

# Define the default storage that’s used if a specified table doesn’t have any table mapping
scalar.db.multi_storage.default_storage=storage0
```

charts/scalardb/templates/scalardb/deployment.yaml
```diff
@@ -72,6 +73,11 @@ spec:
             value: "{{ .Values.scalardb.storageConfiguration.storage }}"
           - name: SCALAR_DB_LOG_LEVEL
             value: "{{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
+          {{- if .Values.scalardb.secretName }}
+          envFrom:
```
ymorimo, line 77: What if the same environment variables are defined in `env` and `envFrom`? It seems that their priority is not clear. I googled a little but could not find the specs. Does `envFrom` have precedence over `env` always, or does the order matter?

ymorimo (approved):
LGTM! Thank you!
Left one minor question.

charts/scalardb/templates/scalardb/deployment.yaml
kota2and3kan: Thank you for your review!

Indeed, I am concerned about priorities.

I tried to test using the same (duplicated) environment variable name.
As a result, the values in `env` are set as environment variables in the Pod, even if I changed the order of `env` and `envFrom`.
So, it seems that the priority does not order matter.

However, I don't find the information in the perspective of specification of Kubernetes yet.
I continue to investigate about priority between `env` and `envFrom`.

---

### Current chart

1. Create `Secret` includes duplicated environment variables.
   ```console
   kubectl create secret generic scalardb-customized-secret \
     --from-literal=SCALAR_DB_CONTACT_POINTS=duplicate_test_contact_point \
     --from-literal=SCALAR_DB_CONTACT_PORT=duplicate_test_port \
     --from-literal=SCALAR_DB_USERNAME=duplicate_test_username \
     --from-literal=SCALAR_DB_PASSWORD=duplicate_test_password \
     --from-literal=SCALAR_DB_STORAGE=duplicate_test_storage \
     --from-literal=SCALAR_DB_LOG_LEVEL=duplicate_test_log_level
   ```
   ```yaml
   $ kubectl get secret scalardb-customized-secret -o yaml
   apiVersion: v1
   data:
     SCALAR_DB_CONTACT_POINTS: ZHVwbGljYXRlX3Rlc3RfY29udGFjdF9wb2ludA==
     SCALAR_DB_CONTACT_PORT: ZHVwbGljYXRlX3Rlc3RfcG9ydA==
     SCALAR_DB_LOG_LEVEL: ZHVwbGljYXRlX3Rlc3RfbG9nX2xldmVs
     SCALAR_DB_PASSWORD: ZHVwbGljYXRlX3Rlc3RfcGFzc3dvcmQ=
     SCALAR_DB_STORAGE: ZHVwbGljYXRlX3Rlc3Rfc3RvcmFnZQ==
     SCALAR_DB_USERNAME: ZHVwbGljYXRlX3Rlc3RfdXNlcm5hbWU=
   kind: Secret
   metadata:
     creationTimestamp: "2022-06-07T04:39:23Z"
     name: scalardb-customized-secret
     namespace: default
     resourceVersion: "186180"
     uid: c831b433-d9d0-403f-9c86-109debe6a935
   type: Opaque
   ```
1. Prepare custom values file.
   ```yaml
   scalardb:
   
     image:
       repository: "local/test/scalardb-server"
       tag: "4.0.0-SNAPSHOT"
   
     secretName: "scalardb-customized-secret"
   ```
1. Deploy Scalar DB Server.
   ```console
   helm install hoge ~/github.com/scalar-labs/helm-charts/charts/scalardb/ -f ./scalardb-server.yaml
   ```
1. Confirm `database.properties` file in the Pod.
   ```console
   $ kubectl exec -t hoge-scalardb-78b449c5b6-bt7qr -- cat database.properties | grep -v -e "^\s*#" -e "^\s*$"
   scalar.db.contact_points=cassandra
   scalar.db.contact_port=9042
   scalar.db.username=cassandra
   scalar.db.password=cassandra
   scalar.db.storage=cassandra
   ```
1. Confirm environment variables in the Pod.
   ```console
   $ kubectl exec -t hoge-scalardb-78b449c5b6-bt7qr -- env | grep -e "^SCALAR_DB"
   SCALAR_DB_STORAGE=cassandra
   SCALAR_DB_USERNAME=cassandra
   SCALAR_DB_CONTACT_POINTS=cassandra
   SCALAR_DB_CONTACT_PORT=9042
   SCALAR_DB_LOG_LEVEL=INFO
   SCALAR_DB_PASSWORD=cassandra
   SCALAR_DB_TRANSACTION_MANAGER=consensus-commit
   SCALAR_DB_ISOLATION_LEVEL=
   SCALAR_DB_CONSENSUSCOMMIT_SERIALIZABLE_STRATEGY=
   ```

---

### Move `envFrom` onto `env`

1. Update chart
   ```diff
   $ git diff
   diff --git a/charts/scalardb/templates/scalardb/deployment.yaml b/charts/scalardb/templates/scalardb/deployment.yaml
   index 209b7ec..9928f62 100644
   --- a/charts/scalardb/templates/scalardb/deployment.yaml
   +++ b/charts/scalardb/templates/scalardb/deployment.yaml
   @@ -46,6 +46,11 @@ spec:
              ports:
                - containerPort: 60051
                - containerPort: 8080
   +          {{- if .Values.scalardb.secretName }}
   +          envFrom:
   +          - secretRef:
   +              name: "{{ .Values.scalardb.secretName }}"
   +          {{- end }}
              env:
              - name: SCALAR_DB_CONTACT_POINTS
                value: "{{ .Values.scalardb.storageConfiguration.contactPoints }}"
   @@ -73,11 +78,6 @@ spec:
                value: "{{ .Values.scalardb.storageConfiguration.storage }}"
              - name: SCALAR_DB_LOG_LEVEL
                value: "{{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
   -          {{- if .Values.scalardb.secretName }}
   -          envFrom:
   -          - secretRef:
   -              name: "{{ .Values.scalardb.secretName }}"
   -          {{- end }}
              resources:
                {{- toYaml .Values.scalardb.resources | nindent 12 }}
              livenessProbe:
   ```
1. Prepare custom values file.
   ```yaml
   scalardb:
   
     image:
       repository: "local/test/scalardb-server"
       tag: "4.0.0-SNAPSHOT"
   
     secretName: "scalardb-customized-secret"
   ```
1. Confirm `Secret`.
   ```yaml
   $ kubectl get secret scalardb-customized-secret -o yaml
   apiVersion: v1
   data:
     SCALAR_DB_CONTACT_POINTS: ZHVwbGljYXRlX3Rlc3RfY29udGFjdF9wb2ludA==
     SCALAR_DB_CONTACT_PORT: ZHVwbGljYXRlX3Rlc3RfcG9ydA==
     SCALAR_DB_LOG_LEVEL: ZHVwbGljYXRlX3Rlc3RfbG9nX2xldmVs
     SCALAR_DB_PASSWORD: ZHVwbGljYXRlX3Rlc3RfcGFzc3dvcmQ=
     SCALAR_DB_STORAGE: ZHVwbGljYXRlX3Rlc3Rfc3RvcmFnZQ==
     SCALAR_DB_USERNAME: ZHVwbGljYXRlX3Rlc3RfdXNlcm5hbWU=
   kind: Secret
   metadata:
     creationTimestamp: "2022-06-07T04:39:23Z"
     name: scalardb-customized-secret
     namespace: default
     resourceVersion: "186180"
     uid: c831b433-d9d0-403f-9c86-109debe6a935
   type: Opaque
   ```
1. Deploy Scalar DB Server.
   ```console
   helm install fuga ~/github.com/scalar-labs/helm-charts/charts/scalardb/ -f ./scalardb-server.yaml
   ```
1. Confirm `database.properties` file in the Pod.
   ```console
   $ kubectl exec -t fuga-scalardb-86c7fc985d-n7xpx -- cat database.properties | grep -v -e "^\s*#" -e "^\s*$"
   scalar.db.contact_points=cassandra
   scalar.db.contact_port=9042
   scalar.db.username=cassandra
   scalar.db.password=cassandra
   scalar.db.storage=cassandra
   ```
1. Confirm environment variables in the Pod.
   ```console
   $ kubectl exec -t fuga-scalardb-86c7fc985d-n7xpx -- env | grep -e "^SCALAR_DB"
   SCALAR_DB_PASSWORD=cassandra
   SCALAR_DB_STORAGE=cassandra
   SCALAR_DB_USERNAME=cassandra
   SCALAR_DB_CONTACT_POINTS=cassandra
   SCALAR_DB_CONTACT_PORT=9042
   SCALAR_DB_LOG_LEVEL=INFO
   SCALAR_DB_TRANSACTION_MANAGER=consensus-commit
   SCALAR_DB_ISOLATION_LEVEL=
   SCALAR_DB_CONSENSUSCOMMIT_SERIALIZABLE_STRATEGY=
   ```

kota2and3kan: I found the docs about `env` and `envFrom`.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#container-v1-core

* envFrom  
  List of sources to populate environment variables in the container. The keys defined within a source must be a C_IDENTIFIER. All invalid keys will be reported as an event when the container is starting. When a key exists in multiple sources, the value associated with the last source will take precedence. Values defined by an Env with a duplicate key will take precedence. Cannot be updated.

According to this docs, the priority of `env` is high.
If the environment variable names are duplicated between `env` and `envFrom`, the value of `env` will be applied to the Pod.

I will add the note about the above in the document that I'm creating #104.

ymorimo: The environment variables like `SCALAR_DB_CONTACT_POINTS` are always defined (even if `.Values.scalardb.storageConfiguration.contactPoints` is null, for example), so isn't it possible for `envFrom` to override these pre-defined variables?
Perhaps should we wrap each `env` variable with `if` statements?

```
          env:
          {{- if Values.scalardb.storageConfiguration.contactPoints }}
          - name: SCALAR_DB_CONTACT_POINTS
            value: "{{ .Values.scalardb.storageConfiguration.contactPoints }}"
          {{- end }}
```

kota2and3kan:
According to the discussion in the meeting, I updated the `values.yaml` and `README`.
In the 1a7a397d4f0157235a1e40b389a6d65d534a1bd1, I omitted long default value since the long database.properties is difficult to read in the `README`.

charts/scalardb/templates/scalardb/deployment.yaml
ymorimo: > However, the default values are defined in the `values.yaml`.
> So, if we use `{{- if Values.scalardb.storageConfiguration.contactPoints }}` as condition, we need to override its value to null (empty string or 0) explicitly in the custom values file for remove pre-defined environment variables.

I see. I missed that point.

> I think it is better and simple to add restrictions in the document as follows.

That's a good idea.

If we make the variable names reserved, why not change the variable names to be different from the env variables defined in the Dockerfile? Probably they are only used as template variables in the `database.properties` template in the yaml file to substitute the (deprecated) `storageConfiguration` parameters, and they are no longer related to the variables in the Dockerfile.

For example, would it be possible to name these variables as follows? It's easier to understand that the values in the `storageConfiguration` block are substituted in the template since the names are straightforward.

```
    scalar.db.contact_points={{ default .Env.storageConfiguration_contactPoints "" }}
    scalar.db.contact_port={{ default .Env.storageConfiguration_contactPort "" }}
    ...
```

kota2and3kan: > If we make the variable names reserved, why not change the variable names to be different from the env variables defined in the Dockerfile? Probably they are only used as template variables in the `database.properties` template in the yaml file to substitute the (deprecated) `storageConfiguration` parameters, and they are no longer related to the variables in the Dockerfile.

You're right.
These environment variable names are only used in Helm Charts.
And, they are not related to the environment variables in the Dockerfile.

We can separate the environment variable name between Helm Charts and Dockerfile because the Helm Charts completely override the database.properties.

> For example, would it be possible to name these variables as follows? It's easier to understand that the values in the `storageConfiguration` block are substituted in the template since the names are straightforward.

I agree with your suggestion (using Helm Charts original environment variable name).
I will update it. Thank you!

kota2and3kan: @ymorimo  
I updated environment variable names in the `values.yaml` and `deployment.yaml` in 266f2eafb8042f90b21bf7f3036a32b2f75c8382.

Regarding the `scalardb.storageConfiguration.dbLogLevel`, it is NOT passed to the `database.properties`.
It's used in the `log4j.properties.tmpl` in the Dockerfile.
And, the environment variable name (`SCALAR_DB_LOG_LEVEL`) depends on the `log4j.properties.tmpl`.
So, we need to remain the name `SCALAR_DB_LOG_LEVEL` same as environment variable name in the `log4j.properties.tmpl`.

Please take a look again!

---

I confirmed the updated environment variables are passed to the Pod correctly as follows.

* `scalardb-server.yaml`
  ```yaml
  scalardb:
  
    image:
      repository: "local/test/scalardb-server"
      tag: "4.0.0-SNAPSHOT"
  
    storageConfiguration:
      contactPoints: "jdbc:postgresql://postgresql-scalardb.default.svc.cluster.local:5432/postgres"
      username: "postgres"
      password: "postgres"
      storage: "jdbc"
  ```
* `database.properties` in the Pod.
  ```console
  $ kubectl exec -t foo-scalardb-56746689b6-k8rp2 -- cat database.properties | grep -v -e "^\s*#" -e "^\s*$"
  scalar.db.contact_points=jdbc:postgresql://postgresql-scalardb.default.svc.cluster.local:5432/postgres
  scalar.db.contact_port=9042
  scalar.db.username=postgres
  scalar.db.password=postgres
  scalar.db.storage=jdbc
  ```
* Environment variables in the Pod.
  ```console
  $ kubectl exec -t foo-scalardb-56746689b6-k8rp2 -- env | grep storage
  storageConfiguration_contactPort=9042
  storageConfiguration_username=postgres
  storageConfiguration_password=postgres
  storageConfiguration_storage=jdbc
  storageConfiguration_contactPoints=jdbc:postgresql://postgresql-scalardb.default.svc.cluster.local:5432/postgres
  ```

superbrothers: Although it is just customary for environment variable names to be in uppercase, I feel that it is strange for environment variable names to be in lowercase. However, there is no problem.

superbrothers: > We can separate the environment variable name between Helm Charts and Dockerfile because the Helm Charts completely override the database.properties.

I feel it would be easier to understand if the environment variable names were the same as those used in the configuration file included in the container image. I understand that with this change, the configuration file included in the container image will no longer be used, and the configuration file defined here will be used instead. If that is the case, I think it is better to use the same environment variable name as defined in the container image.

However, there is no problem with changing the names of environment variables.

superbrothers (approved):
I have a few comments, but they are non-functional areas. The changes themselves look good.

charts/scalardb/templates/scalardb/deployment.yaml
kota2and3kan: @superbrothers  
Thank you for your comments!
We discussed about environment variable name.
As a result, we decided as follows.

* Use uppercase for environment variable name.
* If the environment variable name (secret key name) of `env` and `envFrom` are the same, the value of `env` will be applied since the priority of `env` is high. So, if we use `SCALAR_DB_XXXX` in the chart, users cannot use `SCALAR_DB_XXXX` as a secret key to pass the credentials since the priority of `env` (internal used `SCALAR_DB_XXXX`) is high. For the reason the above, we want to use another name (original name) in the helm internal.
* These environment variables are used in the Helm internal only, but we (users) can see them in the container after deploying. So, we add the prefix `HELM_` so that users know they are internally used environment variables.
* We need to apply the same updates to another scalar products (Ledger, Auditor, etc...). To make changes in the several charts simple and understandable, we just add prefix `HELM_`.
* We will consider to removing default value of Dockerfile separately in the another PR in the future.

kota2and3kan:
@feeblefakie @ymorimo  
I tested and updated the internal used environment variable name according to the discussion. In my test environment, it seems works well.

Please take a look again!

---

### For your reference

1. Create `Secret` with key name `SCALAR_DB_XXXX`
   ```console
   kubectl create secret generic scalardb-customized-secret \
     --from-literal=SCALAR_DB_CONTACT_POINTS="jdbc:postgresql://postgresql-scalardb.default.svc.cluster.local:5432/postgres" \
     --from-literal=SCALAR_DB_CONTACT_PORT=5432 \
     --from-literal=SCALAR_DB_USERNAME=postgres \
     --from-literal=SCALAR_DB_PASSWORD=xxxxxxxx \
     --from-literal=SCALAR_DB_STORAGE=jdbc
   ```
1. Deploy Scalar DB Server with the following custom values file.
   ```yaml
   scalardb:
   
     image:
       repository: "local/test/scalardb-server"
       tag: "4.0.0-SNAPSHOT"
   
     databaseProperties: |
       scalar.db.contact_points={{ default .Env.SCALAR_DB_CONTACT_POINTS "" }}
       scalar.db.contact_port={{ default .Env.SCALAR_DB_CONTACT_PORT "" }}
       scalar.db.storage={{ default .Env.SCALAR_DB_STORAGE "" }}
   
     secretName: "scalardb-customized-secret"
   ```
1. Confirm the environment variable in the container.
   ```console
   $ kubectl exec -it scalardb-c87fc94fb-hbjzp -- env | grep -e "^HELM_"
   HELM_SCALAR_DB_PASSWORD=cassandra
   HELM_SCALAR_DB_STORAGE=cassandra
   HELM_SCALAR_DB_CONTACT_POINTS=cassandra
   HELM_SCALAR_DB_CONTACT_PORT=9042
   HELM_SCALAR_DB_USERNAME=cassandra
   ```
   ```console
   $ kubectl exec -it scalardb-c87fc94fb-hbjzp -- env | grep -e "^SCALAR_DB_"
   SCALAR_DB_USERNAME=postgres
   SCALAR_DB_LOG_LEVEL=INFO
   SCALAR_DB_PASSWORD=xxxxxxxx
   SCALAR_DB_CONTACT_PORT=5432
   SCALAR_DB_STORAGE=jdbc
   SCALAR_DB_CONTACT_POINTS=jdbc:postgresql://postgresql-scalardb.default.svc.cluster.local:5432/postgres
   SCALAR_DB_TRANSACTION_MANAGER=consensus-commit
   SCALAR_DB_ISOLATION_LEVEL=
   SCALAR_DB_CONSENSUSCOMMIT_SERIALIZABLE_STRATEGY=
   ```
1. Confirm the database.properties file in the container.
   ```console
   $ kubectl exec -it scalardb-c87fc94fb-hbjzp -- cat database.properties
   scalar.db.contact_points=jdbc:postgresql://postgresql-scalardb.default.svc.cluster.local:5432/postgres
   scalar.db.contact_port=5432
   scalar.db.username=postgres
   scalar.db.password=xxxxxxxx
   scalar.db.storage=jdbc
   ```

As the above, we can use `SCALAR_DB_XXXX` to pass the credentials via `Secret` resource.

feeblefakie (approved):
LGTM! Thank you!

ymorimo (approved):
Thank you for the discussion! LGTM!

PR #98, opened 2022-05-25
Use Secret resource for username and password in the Scalar DB chart

kota2and3kan:
This PR updates Scalar DB Helm Charts to use `Secret` resource for passing username/password to containers.

As a result of discussion, we decided to update Scalar DB Dockerfile and Helm Charts to the same as Scalar DL for keeping consistency among Scalar products and for improving maintainability.
So, I update Scalar DB Helm Charts based on the Scalar DL Helm Charts in this PR.

Please take a look!

---

Note that we need to apply this update after releasing new version Scalar DB that includes the following PR.
* https://github.com/scalar-labs/scalardb/pull/590

So, we cannot test this Helm Charts now (CI does not work).
We can only test it in the local environment only.
For example, you can test this chart on minikube as follows.

* Create a docker image of Scalar DB Server that supports templating by docerize command.
  ```
  git clone https://github.com/scalar-labs/scalardb.git
  cd scalardb/
  git checkout -b update-dockerfile-to-use-dockerize origin/update-dockerfile-to-use-dockerize
  eval $(minikube -p minikube docker-env)
  ./gradlew docker
  docker tag ghcr.io/scalar-labs/scalardb-server:4.0.0-SNAPSHOT local/test/scalardb-server:4.0.0-SNAPSHOT
  ```
* Run a cassandra container
  ```
  helm install cassandra bitnami/cassandra --set dbUser.password=cassandra
  ```
* Create custom values file
  ```
  cat << EOF > test-custom-values.yaml
  scalardb:
    image:
      repository: "local/test/scalardb-server"
      tag: "4.0.0-SNAPSHOT"
  EOF
  ```
* Deploy Scalar DB using Helm
  ```
  git clone https://github.com/scalar-labs/helm-charts.git
  cd helm-charts/
  git checkout -b use-secret-for-password-scalardb origin/use-secret-for-password-scalardb
  helm install scalardb ./charts/scalardb/ -f /path/to/test-custom-values.yaml
  ```

---

As I mentioned above, since the CI does not work now, I confirmed the following points in my local environment.

* Environment variables in the Scalar DB Server Pod
  ```
  $ kubectl describe pod scalardb-57c8784fc8-lszkh
  Name:         scalardb-57c8784fc8-lszkh
  
  (snip)
  
      Environment:
        SCALAR_DB_CONTACT_POINTS:  cassandra
        SCALAR_DB_CONTACT_PORT:    9042
        SCALAR_DB_USERNAME:        <set to the key 'db-username' in secret 'scalardb-secret'>  Optional: false
        SCALAR_DB_PASSWORD:        <set to the key 'db-password' in secret 'scalardb-secret'>  Optional: false
        SCALAR_DB_STORAGE:         cassandra
        SCALAR_DB_LOG_LEVEL:       INFO
        SCALAR_DB_SERVER_PORT:     50051
  ```
  * We can see the environment variables and the default values are set based on the values.yaml.
* Secret resource
  ```
  $ kubectl get secret scalardb-secret -o jsonpath='{.data.db-username}' | base64 -d
  cassandra
  
  $ kubectl get secret scalardb-secret -o jsonpath='{.data.db-password}' | base64 -d
  cassandra
  ```
  * We can see the Secret resource and the default values are set based on the values.yaml.
* `database.properties` file in the Pod
  ```
  $ kubectl exec -it scalardb-57c8784fc8-lszkh -- cat database.properties | grep -v -e "^\s*#" -e "^\s*$"
  scalar.db.contact_points=cassandra
  scalar.db.contact_port=9042
  scalar.db.username=cassandra
  scalar.db.password=cassandra
  scalar.db.storage=cassandra
  scalar.db.transaction_manager=consensus-commit
  scalar.db.consensus_commit.isolation_level=
  scalar.db.isolation_level=
  scalar.db.consensus_commit.serializable_strategy=
  scalar.db.server.port=50051
  ```
  * The default values are set in the `database.properties` in the Scalar DB Server container. This file created by the `dockerize` command using environment variables.

### DIFF
--- a/charts/scalardb/templates/scalardb/configmap.yaml
+++ b/charts/scalardb/templates/scalardb/configmap.yaml
@@ -1,18 +1,3 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: {{ include "scalardb.fullname" . }}-config-file
-  namespace: {{ .Release.Namespace }}
-data:
-  # Create a database.properties file which is config file of Scalar DB server.
-  database.properties: |-
-    scalar.db.contact_points={{.Values.scalardb.storageConfiguration.contactPoints}}
-    scalar.db.contact_port={{.Values.scalardb.storageConfiguration.contactPort}}
-    scalar.db.username={{.Values.scalardb.storageConfiguration.username}}
-    scalar.db.password={{.Values.scalardb.storageConfiguration.password}}
-    scalar.db.storage={{.Values.scalardb.storageConfiguration.storage}}
-    scalar.db.server.port=50051
----
 {{- if .Values.scalardb.grafanaDashboard.enabled }}
 apiVersion: v1
 kind: ConfigMap
--- a/charts/scalardb/templates/scalardb/deployment.yaml
+++ b/charts/scalardb/templates/scalardb/deployment.yaml
@@ -36,7 +36,6 @@ spec:
       {{- end }}
       securityContext:
         {{- toYaml .Values.scalardb.podSecurityContext | nindent 8 }}
-
       containers:
         - name: {{ .Chart.Name }}
           securityContext:
@@ -47,8 +46,34 @@ spec:
             - containerPort: 50051
             - containerPort: 8080
           env:
-          - name: JAVA_OPTS
-            value: "-Dlog4j.logLevel={{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
+          - name: SCALAR_DB_CONTACT_POINTS
+            value: "{{ .Values.scalardb.storageConfiguration.contactPoints }}"
+          - name: SCALAR_DB_CONTACT_PORT
+            value: "{{ .Values.scalardb.storageConfiguration.contactPort }}"
+          - name: SCALAR_DB_USERNAME
+            valueFrom:
+              secretKeyRef:
+              {{- if .Values.scalardb.existingSecret }}
+                name: {{ .Values.scalardb.existingSecret }}
+              {{- else }}
+                name: {{ include "scalardb.fullname" . }}-secret
+              {{- end }}
+                key: db-username
+          - name: SCALAR_DB_PASSWORD
+            valueFrom:
+              secretKeyRef:
+              {{- if .Values.scalardb.existingSecret }}
+                name: {{ .Values.scalardb.existingSecret }}
+              {{- else }}
+                name: {{ include "scalardb.fullname" . }}-secret
+              {{- end }}
+                key: db-password
+          - name: SCALAR_DB_STORAGE
+            value: "{{ .Values.scalardb.storageConfiguration.storage }}"
+          - name: SCALAR_DB_LOG_LEVEL
+            value: "{{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
+          - name: SCALAR_DB_SERVER_PORT
+            value: "50051"
           resources:
             {{- toYaml .Values.scalardb.resources | nindent 12 }}
           livenessProbe:
@@ -61,14 +86,6 @@ spec:
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1
-          volumeMounts:
-            - name: scalardb-config-file-volume
-              mountPath: /scalardb/server/database.properties
-              subPath: database.properties
-      volumes:
-        - name: scalardb-config-file-volume
-          configMap:
-            name: {{ include "scalardb.fullname" . }}-config-file
       {{- with .Values.scalardb.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/scalardb/templates/scalardb/secret.yaml
+++ b/charts/scalardb/templates/scalardb/secret.yaml
@@ -1,0 +1,13 @@
+{{- if not .Values.scalardb.existingSecret }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "scalardb.fullname" . }}-secret
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "scalardb.labels" . | nindent 4 }}
+type: Opaque
+data:
+  db-username: {{ .Values.scalardb.storageConfiguration.username | b64enc | quote }}
+  db-password: {{ .Values.scalardb.storageConfiguration.password | b64enc | quote }}
+{{- end -}}


charts/scalardb/templates/scalardb/configmap.yaml
```diff
@@ -1,18 +1,3 @@
...
-data:
-  # Create a database.properties file which is config file of Scalar DB server.
-  database.properties: |-
-    scalar.db.contact_points={{.Values.scalardb.storageConfiguration.contactPoints}}
-    scalar.db.contact_port={{.Values.scalardb.storageConfiguration.contactPort}}
-    scalar.db.username={{.Values.scalardb.storageConfiguration.username}}
-    scalar.db.password={{.Values.scalardb.storageConfiguration.password}}
-    scalar.db.storage={{.Values.scalardb.storageConfiguration.storage}}
-    scalar.db.server.port=50051
----
```
kota2and3kan, line 15: We pass all configurations via environment variables. So, I deleted the `ConfigMap`.

charts/scalardb/templates/scalardb/deployment.yaml
```diff
@@ -47,8 +46,34 @@ spec:
             - containerPort: 50051
             - containerPort: 8080
           env:
-          - name: JAVA_OPTS
-            value: "-Dlog4j.logLevel={{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
```
kota2and3kan, line 51: After this update, we need to specify the log level using `SCALAR_DB_LOG_LEVEL` instead of `JAVA_OPTS`.

charts/scalardb/templates/scalardb/deployment.yaml
```diff
@@ -47,8 +46,34 @@ spec:
...
+              {{- else }}
+                name: {{ include "scalardb.fullname" . }}-secret
+              {{- end }}
+                key: db-password
+          - name: SCALAR_DB_STORAGE
+            value: "{{ .Values.scalardb.storageConfiguration.storage }}"
+          - name: SCALAR_DB_LOG_LEVEL
+            value: "{{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
+          - name: SCALAR_DB_SERVER_PORT
+            value: "50051"
```
kota2and3kan, line 76: In the current chart, the value of `scalar.db.server.port` is hard-coded as `50051`.
This port is only used internal connections and is not exposed to user to access Scalar DB Server.
Basically, uses accesss Scalar DB Server via Service resource (Envoy).

So, I think that this hard-coding is not critical issue and we don't need fix now.
However, if we need to fix it in the future, I will do that in the another PR.

ymorimo (approved):
LGTM! Thank you!
I will update #89 accordingly.

charts/scalardb/templates/scalardb/deployment.yaml
ymorimo: I have just noticed that the default port of the Scalar DB Server is 60051, isn't it?
And if we use the default port, perhaps is it possible to leave the value blank in this yaml file?

kota2and3kan: Yes. The default port of Scalar DB Server is 60051.
I think this configuration (hard-coded 50051) related to the Envoy. 
It's just a guess, but I think the default values of Scalar Envoy Charts (50051 as target port) is set based on the assumption of Scalar DL.

So, I think we can use 60051 port and remove hard-coded value, if I we set target port of Envoy to 60051 and fix several hard-coded point (50051) in the YAML file.

However, I didn't test the above configuration yet.
So, I will try to test using above configurations today.
Could you please give me some time?

ymorimo: Got it! Let's keep the port 50051 for now. Thank you for confirming!

kota2and3kan: For your reference.

I checked the configurations of Scalar Envoy, Scalar Envoy Charts, and Scalar DB Charts.
As a result, I found the way to remove this hard-coded value.
I need to change the `deployment.yaml` and to use custom values file as follows.

* File changed
  ```diff
  --- a/charts/scalardb/templates/scalardb/deployment.yaml
  +++ b/charts/scalardb/templates/scalardb/deployment.yaml
  @@ -43,7 +43,7 @@ spec:
             image: "{{ .Values.scalardb.image.repository }}:{{ .Values.scalardb.image.tag | default .Chart.AppVersion }}"
             imagePullPolicy: {{ .Values.scalardb.image.pullPolicy }}
             ports:
  -            - containerPort: 50051
  +            - containerPort: 60051
               - containerPort: 8080
             env:
             - name: SCALAR_DB_CONTACT_POINTS
  @@ -72,15 +72,13 @@ spec:
               value: "{{ .Values.scalardb.storageConfiguration.storage }}"
             - name: SCALAR_DB_LOG_LEVEL
               value: "{{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
  -          - name: SCALAR_DB_SERVER_PORT
  -            value: "50051"
             resources:
               {{- toYaml .Values.scalardb.resources | nindent 12 }}
             livenessProbe:
               exec:
                 command:
                 - /usr/local/bin/grpc_health_probe
  -              - -addr=:50051
  +              - -addr=:60051
               failureThreshold: 3
               initialDelaySeconds: 10
               periodSeconds: 10
  ```
* Custom values file
  ```yaml
  envoy:
    envoyConfiguration:
      serviceListeners: "scalardb-service:60051"
    service:
      ports: 
        envoy:
          port: 60051
          targetPort: 60051
      type: "LoadBalancer" # For testing. This is not necessary.
  scalardb:
    image:
      repository: "local/test/scalardb-server"
      tag: "4.0.0-SNAPSHOT"
    service:
      ports: 
        scalardb:
          port: 60051
          targetPort: 60051
  ```

When I apply the above changes and configuration, I was able to run Scalar DB Server with default port (60051).
I think we should change the default values (values.yaml file), if we want to remove this hard-coded value.

However, it works well for now, and I think this changes is not necessary for running Scalar DB Server using Helm Charts.
So, I will discuss and consider to fix this hard-coded value as an another task in the future.

feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@superbrothers Merging it but PTAL when you get a chance!

superbrothers:
LGTM!

PR #100, opened 2022-05-30
Use default port in the chart of Scalar DB and Auditor

kota2and3kan:
This PR updates Scalar DB and Auditor charts to use default ports (Scalar DB -> 60051 / Auditor -> 40051, 40052).

Please take a look!

### DIFF
--- a/charts/scalardb/templates/scalardb/deployment.yaml
+++ b/charts/scalardb/templates/scalardb/deployment.yaml
@@ -43,7 +43,7 @@ spec:
           image: "{{ .Values.scalardb.image.repository }}:{{ .Values.scalardb.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.scalardb.image.pullPolicy }}
           ports:
-            - containerPort: 50051
+            - containerPort: 60051
             - containerPort: 8080
           env:
           - name: SCALAR_DB_CONTACT_POINTS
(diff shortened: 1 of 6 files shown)


charts/scalardb/templates/scalardb/deployment.yaml
```diff
@@ -72,15 +72,13 @@ spec:
             value: "{{ .Values.scalardb.storageConfiguration.storage }}"
           - name: SCALAR_DB_LOG_LEVEL
             value: "{{ .Values.scalardb.storageConfiguration.dbLogLevel }}"
-          - name: SCALAR_DB_SERVER_PORT
-            value: "50051"
```
kota2and3kan, line 76: When we use default port (60051) in the Scalar DB Helm Charts, we can remove this hard-coded value from manifest file.

charts/scalardb/values.yaml
```diff
@@ -19,7 +19,7 @@ envoy:
     # -- admin log path
     adminAccessLogPath: /dev/stdout
     # -- list of service name and port
-    serviceListeners: scalardb-service:50051,scalardb-privileged:50052
+    serviceListeners: scalardb-service:60051
```
kota2and3kan, line 22: Scalar DB doesn't have **privileged port** like Scalar DL.
So, we need the `scalardb-service:60051` only.

ymorimo (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #101, opened 2022-05-30
Remove unused admin port from Auditor chart

kota2and3kan:
This PR removes unused admin port (50053) from Scalar DL Auditor chart.


kota2and3kan:
From the Scalar DL v3.5.0, Auditor use admin port.
And, I updated Auditor chart to use `40053` port for scalar-admin by default in #124.
So, I close this PR.

PR #102, opened 2022-06-01
Support customized ledger.properties in Scalar DL Ledger

kota2and3kan:
This PR add a new feature that accepting user customized ledger.properties file in the Scalar DL Ledger Helm Charts.
By default, it is disabled.

Please take a look!

---

## How to use your customized ledger.properties file

You can use your customized ledger.properties by the following steps.

1. Set `ledger.useCustomizedConfiguration.enabled=true` in your custom value file.
1. Create a `ConfigMap` resource using your ledger.properties file as follows.
   ```console
   kubectl create configmap ledger-customized-config --from-file=ledger.properties.tmpl=ledger.properties
   ```
1. Create schema using the Scalar DL Schema Loader Helm Chats.
1. Deploy Scalar DL using Helm Charts.

---

## How to pass the credentials as a environment variables using `Secret` resource

Also, you can use a `Secret` resource to pass the credentials as a environment variables as follows.

1. Set `ledger.useCustomizedConfiguration.useSecret=true` in your custom value file.
1. Create the `ConfigMap` includes environment variable name as Go template syntax.
   ```
   scalar.dl.ledger.proof.enabled=false
   scalar.dl.ledger.auditor.enabled=false
   
   scalar.db.contact_points=jdbc:postgresql://postgresql-ledger:5432/postgres
   scalar.db.username={{ default .Env.CUSTOMIZED_CONFIG_POSTGRES_USERNAME "" }}
   scalar.db.password={{ default .Env.CUSTOMIZED_CONFIG_POSTGRES_PASSWORD "" }}
   scalar.db.storage=jdbc
   ```
1. Create the `Secret` includes credentials.
   ```console
   kubectl create secret generic ledger-customized-secret \
     --from-literal=customized-config-username=postgres \
     --from-literal=customized-config-password=xxxxxxxx
   ```
1. Specify the **environment variable name** and **key name** of `Secret` in the custom values file.
   ```yaml
   ledger:
     useCustomizedConfiguration:
       enabled: true
       configMapName: "ledger-customized-config"
       useSecret: true
       secretKeys:
         - environmentVariableName: CUSTOMIZED_CONFIG_POSTGRES_USERNAME
           secretKeyName: customized-config-username
         - environmentVariableName: CUSTOMIZED_CONFIG_POSTGRES_PASSWORD
           secretKeyName: customized-config-password
   ```

As a result, the credentials are set in the ledger.properties file in each container as follows.
```
scalar.dl.ledger.proof.enabled=false
scalar.dl.ledger.auditor.enabled=false

scalar.db.contact_points=jdbc:postgresql://postgresql-ledger:5432/postgres
scalar.db.username=postgres
scalar.db.password=xxxxxxxx
scalar.db.storage=jdbc
```

### DIFF
--- a/charts/scalardl/templates/ledger/deployment.yaml
+++ b/charts/scalardl/templates/ledger/deployment.yaml
@@ -34,6 +34,11 @@ spec:
           secret:
             secretName: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
       {{- end }}
+      {{- if .Values.ledger.useCustomizedConfiguration.enabled }}
+        - name: ledger-customized-config-file-volume
+          configMap:
+            name: {{ .Values.ledger.useCustomizedConfiguration.configMapName }}
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}-ledger
           securityContext:
@@ -46,11 +51,17 @@ spec:
               mountPath: "/keys"
               readOnly: true
           {{- end }}
+          {{- if .Values.ledger.useCustomizedConfiguration.enabled }}
+            - name: ledger-customized-config-file-volume
+              mountPath: /scalar/ledger/ledger.properties.tmpl
+              subPath: ledger.properties.tmpl
+          {{- end }}
           ports:
           - containerPort: 50051
           - containerPort: 50052
           - containerPort: 50053
           - containerPort: 8080
+          {{- if not .Values.ledger.useCustomizedConfiguration.enabled }}
           env:
           - name: SCALAR_DB_CONTACT_POINTS
             value: "{{ .Values.ledger.scalarLedgerConfiguration.dbContactPoints }}"
@@ -85,6 +96,16 @@ spec:
           {{- if .Values.ledger.scalarLedgerConfiguration.ledgerAuditorEnabled }}
           - name: SCALAR_DL_LEDGER_AUDITOR_ENABLED
             value: "{{ .Values.ledger.scalarLedgerConfiguration.ledgerAuditorEnabled }}"
+          {{- end }}
+          {{- else if .Values.ledger.useCustomizedConfiguration.useSecret }}
+          env:
+          {{- range .Values.ledger.useCustomizedConfiguration.secretKeys }}
+          - name: {{ .environmentVariableName }}
+            valueFrom:
+              secretKeyRef:
+                name: {{ $.Values.ledger.useCustomizedConfiguration.secretName }}
+                key: {{ .secretKeyName }}
+          {{- end }}
           {{- end }}
           livenessProbe:
             exec:
--- a/charts/scalardl/templates/ledger/deployment.yaml
+++ b/charts/scalardl/templates/ledger/deployment.yaml
@@ -39,7 +39,7 @@ spec:
         - name: scalardl-ledger-properties-volume
           configMap:
             name: {{ include "scalardl.fullname" . }}-ledger-properties
-      {{- with .Values.ledger.volumes }}
+      {{- with .Values.ledger.extraVolumes }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
       containers:
@@ -57,7 +57,7 @@ spec:
             - name: scalardl-ledger-properties-volume
               mountPath: /scalar/ledger/ledger.properties.tmpl
               subPath: ledger.properties.tmpl
-          {{- with .Values.ledger.volumeMounts }}
+          {{- with .Values.ledger.extraVolumeMounts }}
             {{- toYaml . | nindent 12 }}
           {{- end }}
           ports:
--- a/charts/scalardl/templates/ledger/deployment.yaml
+++ b/charts/scalardl/templates/ledger/deployment.yaml
@@ -31,7 +31,7 @@ spec:
       securityContext:
         {{- toYaml .Values.ledger.podSecurityContext | nindent 8 }}
       volumes:
-      {{- if (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
+      {{- if or (.Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled) (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
         - name: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
           secret:
             secretName: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
@@ -46,7 +46,7 @@ spec:
           image: "{{ .Values.ledger.image.repository }}:{{ .Values.ledger.image.version }}"
           imagePullPolicy: {{ .Values.ledger.image.pullPolicy }}
           volumeMounts:
-          {{- if (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
+          {{- if or (.Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled) (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
             - name: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
               mountPath: "/keys"
               readOnly: true
--- a/charts/scalardl/templates/ledger/deployment.yaml
+++ b/charts/scalardl/templates/ledger/deployment.yaml
@@ -17,6 +17,8 @@ spec:
   {{- end }}
   template:
     metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/ledger/configmap.yaml") . | sha256sum }}
       labels:
         {{- include "scalardl-ledger.selectorLabels" . | nindent 8 }}
     spec:
@@ -29,16 +31,14 @@ spec:
       securityContext:
         {{- toYaml .Values.ledger.podSecurityContext | nindent 8 }}
       volumes:
-      {{- if .Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled }}
+      {{- if (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
         - name: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
           secret:
             secretName: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
       {{- end }}
-      {{- if .Values.ledger.useCustomizedConfiguration.enabled }}
-        - name: ledger-customized-config-file-volume
+        - name: scalardl-ledger-properties-volume
           configMap:
-            name: {{ .Values.ledger.useCustomizedConfiguration.configMapName }}
-      {{- end }}
+            name: {{ include "scalardl.fullname" . }}-ledger-properties
       containers:
         - name: {{ .Chart.Name }}-ledger
           securityContext:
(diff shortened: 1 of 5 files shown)


kota2and3kan:
I will update this PR based on the #96 after it will be merged.

charts/scalardl/templates/ledger/deployment.yaml
```diff
@@ -29,34 +31,40 @@ spec:
       securityContext:
         {{- toYaml .Values.ledger.podSecurityContext | nindent 8 }}
       volumes:
-      {{- if .Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled }}
+      {{- if or (.Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled) (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
```
kota2and3kan, line 34: In the Ledger charts, we need to add new condition (if statement) to mount the key file that will be set to `scalar.dl.ledger.proof.private_key_path`.

If we use customized `ledger.properties`, we don't set `ledger.scalarLedgerConfiguration.ledgerProofEnabled` in the custom values file. We set `scalar.dl.ledger.proof.enabled=true` in the `ledger.ledgerProperties` directly instead of `ledger.scalarLedgerConfiguration.ledgerProofEnabled`.

However, in the above case, the following condition returns `false` and the key file is not mounted to the Pod even if we enable ledger proof.

```
{{- if .Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled }}
```

We can avoid this issue by using the following custom values file.

```yaml
ledger:

  scalarLedgerConfiguration:
    ledgerProofEnabled: true

  ledgerProperties: |
    ...
    scalar.dl.ledger.proof.enabled=true
    ...
```

However, I think this way is not good since this custom values file has redundancy settings. It set `scalar.dl.ledger.proof.enabled=true` twice.
(Also, we will face the same issue when we will deprecate the `ledger.scalarLedgerConfiguration` in the future.)

So, I think it is better to use another way to avoid this issue.
And, to resolve this issue, I chose [lookup](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function) function of Helm.

The `lookup` function can be used to look up resources in a running Kubernetes cluster.
The following `if` statement checks the specific named `Secret` resource exists or not.
(`Secret` name is set in the custom values file or values.yaml file.)

```
{{- if (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
```

If the specified named `Secret` (includes key file for Ledger) exists, this chart mounts it to the Pod.
If the specified named `Secret` does NOT exist, this chart doesn't mount anything. (If we set `scalar.dl.ledger.proof.enabled=true` and there is no `Secret`, the Ledger pods will be failed.)

Also, to keep backward-compatibility, I leave the old condition and combine these two conditions using `or` operator.

charts/scalardl/templates/ledger/deployment.yaml
```diff
@@ -74,18 +82,23 @@ spec:
                 name: {{ include "scalardl.fullname" . }}-ledger
               {{- end }}
                 key: db-password
-          - name: SCALAR_DB_STORAGE
+          - name: HELM_SCALAR_DB_STORAGE
             value: "{{ .Values.ledger.scalarLedgerConfiguration.dbStorage }}"
           - name: SCALAR_DL_LEDGER_LOG_LEVEL
```
kota2and3kan, line 87: We cannot add `HELM_` prefix to this environment variable name since this is used in the `log4j.properties` in the container and we cannot override `log4j.properties` using Helm Charts at this time.

kota2and3kan:
@feeblefakie @ymorimo  
I updated this PR based on the Scalar DB chart (#96).
Please take a look!

---

### How to use your customized ledger.properties file

You can use your customized `ledger.properties` by the following steps.

1. Set your `ledger.properties` as a value of `ledger.ledgerProperties` in a custom values file.
   * Example
     ```yaml
     ledger:
     
       ledgerProperties: |
         scalar.db.contact_points=jdbc:postgresql://postgresql-ledger.default.svc.cluster.local:5432/postgres
         scalar.db.username=postgres
         scalar.db.password=xxxxxxxx
         scalar.db.storage=jdbc
         scalar.dl.ledger.proof.enabled=true
         scalar.dl.ledger.auditor.enabled=true
         scalar.dl.ledger.proof.private_key_path=/keys/private-key
     ```
1. Create schema using the Scalar DL Schema Loader Helm Chats.
1. Deploy Scalar DL using Helm Charts.

---

### How to pass the credentials as a environment variables using `Secret` resource

Also, you can use a `Secret` resource to pass the credentials as a environment variables as follows.

1. Set environment variable name to the `ledger.ledgerProperties` in a custom values file using Go template syntax.
   * Example
     ```yaml
     ledger:
     
       ledgerProperties: |
         scalar.db.contact_points=jdbc:postgresql://postgresql-ledger.default.svc.cluster.local:5432/postgres
         scalar.db.storage=jdbc
         scalar.dl.ledger.proof.enabled=true
         scalar.dl.ledger.auditor.enabled=true
         scalar.dl.ledger.proof.private_key_path=/keys/private-key
     
       secretName: "scalardl-customized-secret"
     ```
1. Create `Secret` resource that includes credentials. You need to specify the environment variable name as keys of `Secret`.
   * Example
     ```console
     kubectl create secret generic scalardl-customized-secret \
       --from-literal=SCALAR_DB_USERNAME=postgres \
       --from-literal=SCALAR_DB_PASSWORD=xxxxxxxx
     ```
1. Set the `Secret` name as a value of `ledger.secretName` in the custom values file.
1. Create schema using the Scalar DL Schema Loader Helm Chats.
1. Deploy Scalar DL using Helm Charts.

As a result, the credentials are set in the ledger.properties file in each container as follows.

```console
scalar.db.contact_points=jdbc:postgresql://postgresql-ledger.default.svc.cluster.local:5432/postgres
scalar.db.username=postgres
scalar.db.password=xxxxxxxx
scalar.db.storage=jdbc
scalar.dl.ledger.proof.enabled=true
scalar.dl.ledger.auditor.enabled=true
scalar.dl.ledger.proof.private_key_path=/keys/private-key
```

charts/scalardl/templates/ledger/deployment.yaml
superbrothers: If customized ledger.properties is used, a modified key path may be specified as follows:

```
scalar.dl.ledger.proof.enabled=true
scalar.dl.ledger.proof.private_key_path=/tmp/private.key
```

It seems simpler to have `.Values.ledger.scalarLedgerConfiguration` setting only work with the default `ledger.properties`. In your suggestion above, although you are using customized ledger.properties, the Secret object name needs to match `.Values.ledger.scalarLedgerConfiguration.secretName` and the key name must be `private-key`.

For example, how about allowing volumes and volumeMounts to be customizable and have them used as shown below? This way, if in the future you decide to use credentials other than keys as files in customized ledger.properties, you can do so without changing the chart.

```diff
diff --git a/charts/scalardl/templates/ledger/deployment.yaml b/charts/scalardl/templates/ledger/deployment.yaml
index 88b28d7..6c031af 100644
--- a/charts/scalardl/templates/ledger/deployment.yaml
+++ b/charts/scalardl/templates/ledger/deployment.yaml
@@ -31,7 +31,7 @@ spec:
       securityContext:
         {{- toYaml .Values.ledger.podSecurityContext | nindent 8 }}
       volumes:
-      {{- if or (.Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled) (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
+      {{- if .Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled }}
         - name: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
           secret:
             secretName: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
@@ -39,6 +39,9 @@ spec:
         - name: scalardl-ledger-properties-volume
           configMap:
             name: {{ include "scalardl.fullname" . }}-ledger-properties
+        {{- with .Values.ledger.volumes }}
+          {{- toYaml . nindent 8 }}
+        {{- end }}
       containers:
         - name: {{ .Chart.Name }}-ledger
           securityContext:
@@ -46,7 +49,7 @@ spec:
           image: "{{ .Values.ledger.image.repository }}:{{ .Values.ledger.image.version }}"
           imagePullPolicy: {{ .Values.ledger.image.pullPolicy }}
           volumeMounts:
-          {{- if or (.Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled) (lookup "v1" "Secret" .Release.Namespace .Values.ledger.scalarLedgerConfiguration.secretName) }}
+          {{- if .Values.ledger.scalarLedgerConfiguration.ledgerProofEnabled }}
             - name: "{{ .Values.ledger.scalarLedgerConfiguration.secretName }}"
               mountPath: "/keys"
               readOnly: true
@@ -54,6 +57,9 @@ spec:
             - name: scalardl-ledger-properties-volume
               mountPath: /scalar/ledger/ledger.properties.tmpl
               subPath: ledger.properties.tmpl
+            {{- with .Values.ledger.volumeMounts }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
           ports:
           - containerPort: 50051
           - containerPort: 50052
diff --git a/charts/scalardl/values.yaml b/charts/scalardl/values.yaml
index 3dcf184..8096758 100644
--- a/charts/scalardl/values.yaml
+++ b/charts/scalardl/values.yaml
@@ -210,6 +210,9 @@ ledger:
     # runAsNonRoot: true
     # runAsUser: 1000

+  volumeMounts: []
+  volumes: []
+
   service:
     # -- service types in kubernetes
     type: ClusterIP
```

charts/scalardl/values.yaml
```diff
@@ -173,6 +210,11 @@ ledger:
     # runAsNonRoot: true
     # runAsUser: 1000
 
+  # -- If you set "scalar.dl.ledger.proof.enabled=true" to ledger.ledgerProperties, you need to create volume that includes key file.
+  volumes: []
+  # -- If you set "scalar.dl.ledger.proof.enabled=true" to ledger.ledgerProperties, you need to mount key file to the path set in the "scalar.dl.ledger.proof.private_key_path".
+  volumeMounts: []
```
superbrothers, line 216: [nits] In Grafana Chart, key names seem to be `extraVolumes` and `extraVolumeMounts`. It is indeed easier to understand that they are added.

- https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L457-L467

In addition, the help message mentions `scalar.dl.ledger.proof.enabled`, but the message needs to be updated each time volumes are used outside of the setting. I suggest changing the message to `Defines additional volume mounts` and supplementing the usage with documentation.

charts/scalardl/values.yaml
```diff
@@ -140,6 +140,43 @@ ledger:
     # -- The secret key of a Ledger private key
     ledgerPrivateKeySecretKey: private-key
 
+  # -- The ledger.properties is created based on the values of ledger.scalarLedgerConfiguration by default. If you want to customize ledger.properties, you can override this value with your ledger.properties.
+  # @default -- The minimum template of ledger.properties is set by default.
```
feeblefakie, line 144:
```suggestion
  # @default -- The default minimum necessary values of ledger.properties are set. You can overwrite it with your own ledger.properties.
```

feeblefakie (approved):
LGTM! Thank you!
Left a suggestion about the comment.

charts/scalardl/values.yaml
kota2and3kan: Thank you for the suggestion!
I applied it.

charts/scalardl/values.yaml
kota2and3kan: Thank you for the suggestion!

> In Grafana Chart, key names seem to be `extraVolumes` and `extraVolumeMounts`. It is indeed easier to understand that they are added.

I tried to search similar settings and found some products use `extraVolumes`.

https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml#L542-L547
https://github.com/elastic/helm-charts/blob/main/elasticsearch/values.yaml#L146-L153
https://github.com/hashicorp/consul-k8s/blob/main/charts/consul/values.yaml#L1224

It looks good to me since the name `extra` can imagine its use case.
So, I updated the name to `extraVolumes` and `extraVolumeMounts`.

> In addition, the help message mentions `scalar.dl.ledger.proof.enabled`, but the message needs to be updated each time volumes are used outside of the setting. I suggest changing the message to `Defines additional volume mounts` and supplementing the usage with documentation.

Indeed, it is better to make comment general. There is a possibility that we will have an another use case in the future.
I updated the comment based on the suggestion.
Also, I will explain how to use customized properties and how to mount key files in the document that I'm creating in the following PR.
https://github.com/scalar-labs/helm-charts/pull/104

ymorimo (approved):
LGTM!

superbrothers (approved):

PR #106, opened 2022-06-02
Update chart version to keep main branch latest (Auditor)

kota2and3kan:
A new patch version of Scalar DL Auditor Helm Charts has been released.
This PR updates version of Scalar DL Auditor chart to keep main branch latest.
(This release flow will be fixed in the future.)

Please take a look!

### DIFF
--- a/.git-hooks/pre-commit
+++ b/.git-hooks/pre-commit
@@ -1,0 +1,14 @@
+#! /usr/bin/env bash
+
+#
+# Pre-commit hook for the helm-charts repository
+#
+# Packages and re-indexes the chart repository in the docs directory on every commit.
+#
+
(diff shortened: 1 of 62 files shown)


feeblefakie (approved):
LGTM! Thank you!

I changed the label to `bugfix`. Any reason for the `improvement`?

kota2and3kan:
@feeblefakie  
> I changed the label to bugfix. Any reason for the improvement?

We don't need to apply this patch into the any branch of current/old version.
So, I didn't choose `bugfix`.

And honestly, I didn't find any appropriate tag other than `improvement`.
In other words, I decided by process of elimination.
(I think we don't need to apply this patch in the next minor version too...)

This is the reason I chose `improvement` tag.

PR #108, opened 2022-06-06
Support customized database.properties in values.yaml

kota2and3kan:
This is a PR for discussion.

I implemented the suggested feature in the following comment.
https://github.com/scalar-labs/helm-charts/pull/96#discussion_r889630203

I would like to discuss which way is better between this PR and the following PR.
https://github.com/scalar-labs/helm-charts/pull/96


charts/scalardb/values.yaml
```diff
@@ -123,6 +123,29 @@ scalardb:
     # -- The log level of Scalar DB
     dbLogLevel: INFO
 
+  # -- The database.properties is created based on the values of scalardb.storageConfiguration by default. If you want to customize database.properties, you can override this value with your database.properties.
+  databaseProperties: |
```
kota2and3kan, line 127: This is a main point of this PR.
In this PR, we define database.properties as value of values.yaml.
(In other words, we don't need to create `ConfigMap` manually.)

If we use the default value of `scalardb.databaseProperties`, each value of `scalardb.storageConfiguration` applied to `scalardb.databaseProperties` via environment variables.
(The templating will be done by `dockerize` in the container side.)

Also, if we override this value in a custom values file, the `ConfigMap` is created automatically, and we can use the overridden (customized) database.properties.

charts/scalardb/values.yaml
```diff
@@ -123,6 +123,29 @@ scalardb:
...
+    scalar.db.username={{ default .Env.SCALAR_DB_USERNAME "" }}
+    # The value of scalardb.storageConfiguration.password is set by default.
+    scalar.db.password={{ default .Env.SCALAR_DB_PASSWORD "" }}
+
+    # Storage implementation. "cassandra" or "cosmos" or "dynamo" or "jdbc" or "grpc" can be set. Default storage is "cassandra".
+    # The value of scalardb.storageConfiguration.storage is set by default.
+    scalar.db.storage={{ default .Env.SCALAR_DB_STORAGE "" }}
+
+  # -- Secret name that includes credentials. Each secret key is passed to Pod as environment variables using envFrom.
+  credentialsSecretName: ""
```
kota2and3kan, line 147: If we create the `Secret` resource includes some credentials and specify the `Secret` resource name in this value, each `Secret Key` is passed to Pod as environment variables.

We can use those environment variables for templating of the database.properties.

charts/scalardb/templates/scalardb/deployment.yaml
```diff
@@ -19,6 +19,7 @@ spec:
     metadata:
     {{- if .Values.scalardb.podAnnotations }}
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/scalardb/configmap.yaml") . | sha256sum }}
```
kota2and3kan, line 22: We can detect the changes of database.properties in the custom values file and reload new database.properties (re-create Pods) automatically using this annotation.

If we change database.properties in the custom values file, the `ConfigMap` is changed and the hash value of `ConfigMap` is changed too.
The Helm detects the changes of hash and re-create Pods automatically.

kota2and3kan:
Since the discussion closed, I close this PR.
I will update the Helm Charts in the #96.

PR #110, opened 2022-06-10
Add monitoring resources to scalardb-graphql chart

ymorimo:
This PR adds Service, ServiceMonitor, and PrometheusRule resources to `scalardb-graphql` chart as `scalardb` and `scalardl` charts do.

The latest build of the docker image from the `scalardb-graphql` `main` branch is needed.

I confirmed on a minikube cluster:

* Build a docker image of scalardb-graphql

    ```console
    cd /path/to/scalardb-graphql
    ./gradlew docker
    minikube image load ghcr.io/scalar-labs/scalardb-graphql:4.0.0-SNAPSHOT
    minikube image tag ghcr.io/scalar-labs/scalardb-graphql:4.0.0-SNAPSHOT local/test/scalardb-graphql:4.0.0-SNAPSHOT
    ```

* Start a database in the `minikube` network and load a schema

    ```console
    docker run -d --name cassandra-scalardb --network minikube -p 9042:9042 cassandra:3.11
    java -jar scalardb-schema-loader-3.5.0.jar ...
    ```

* Create a custom values file

    ```
    cat <<EOF > test-custom-values.yaml
    image:
      repository: local/test/scalardb-graphql
      tag: 4.0.0-SNAPSHOT
    
    scalarDbGraphQlConfiguration:
      contactPoints: cassandra-scalardb
      contactPort: 9042
      username: cassandra
      password: cassandra
      storage: cassandra
      logLevel: INFO
      path: /graphql
      namespaces: emoney
      graphiql: "true"
    
    service:
      type: LoadBalancer

    serviceMonitor:
      enabled: true
    prometheusRule:
      enabled: true
    ingress:
      enabled: true
      className: nginx
    EOF
    ```
    
* Deploy using Helm

    ```console
    helm install scalardb-graphql ./charts/scalardb-graphql/ -f /path/to/test-custom-values.yaml
    ```

Then, confirmed that monitoring is working by following the guide: https://github.com/scalar-labs/helm-charts/blob/main/docs/getting-started-monitoring.md

---

The Grafana Dashboard configuration is going to be added in another PR since it will be large.

### DIFF
--- a/charts/scalardb-graphql/templates/scalardb-graphql/service.yaml
+++ b/charts/scalardb-graphql/templates/scalardb-graphql/service.yaml
@@ -16,3 +16,19 @@ spec:
       name: http
   selector:
     {{- include "scalardb-graphql.selectorLabels" . | nindent 4 }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "scalardb-graphql.fullname" . }}-metrics
+  labels:
+    {{- include "scalardb-graphql.labels" . | nindent 4 }}
+spec:
+  ports:
+  - name: scalardb-graphql-prometheus
+    port: 9080
+    targetPort: 9080
+    protocol: TCP
+  selector:
+    {{- include "scalardb-graphql.selectorLabels" . | nindent 4 }}
+  type: ClusterIP
--- a/charts/scalardb-graphql/README.md
+++ b/charts/scalardb-graphql/README.md
@@ -22,6 +22,8 @@ Current chart version is `1.0.0`
 | nameOverride | string | `""` |  |
 | nodeSelector | object | `{}` | nodeSelector is form of node selection constraint. |
 | podSecurityContext | object | `{}` | PodSecurityContext holds pod-level security attributes and common container settings. |
+| prometheusRule.enabled | bool | `false` | enable rules for prometheus |
+| prometheusRule.namespace | string | `"monitoring"` | which namespace prometheus is located. by default monitoring |
 | replicaCount | int | `3` |  |
 | resources | object | `{}` | Resources allowed to the pod. |
 | scalarDbGraphQlConfiguration.consensuscommitIsolationLevel | string | `""` | Default isolation level for ConsensusCommit. |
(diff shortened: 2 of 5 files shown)


charts/scalardb-graphql/templates/scalardb-graphql/service.yaml
```diff
@@ -16,3 +16,19 @@ spec:
...
+kind: Service
+metadata:
+  name: {{ include "scalardb-graphql.fullname" . }}-metrics
+  labels:
+    {{- include "scalardb-graphql.labels" . | nindent 4 }}
+spec:
+  ports:
+  - name: scalardb-graphql-prometheus
+    port: 9080
+    targetPort: 9080
```
ymorimo, line 30: The Prometheus port `9080` differs from `8080` on other charts because `8080` is used as the default port of the GraphQL HTTP server. I feel it was a design mistake. I wonder if we should change the GraphQL HTTP port to another number such as `8000`.

kota2and3kan: This is a minor comment for your reference. This comment does not prevent for merging the PR.

I think the GraphQL HTTP port is used for GraphiQL that is accessed from browsers as follows.

* GraphiQL URL
  ```
  http://localhost:8080/graphql
  ```

Also, the default port of Web UI of Scalar Manager is `8000` now.
The Scalar Manager Web UI is accessed from browsers as follows.
* Scalar Manager Web UI URL
  ```
  http://localhost:8000/
  ```

So, if we change the default port of GraphQL HTTP to `8000`, I think it causes another conflict of port between the GraphiQL and the Scalar Manager Web UI.
Basically, I think both will be exposed to the outside of Kubernetes Cluster to access them from browsers.

I think we need to consider the above thing when we decide the default port.

kota2and3kan (approved):
LGTM! Thank you!

I left a minor comment that is not related to this PR directly.
Please take a look!

brfrn169 (approved):
LGTM! Thank you!

PR #113, opened 2022-07-01
Update default Ingress paths configuration in values.yaml

ymorimo:
This PR updates the default Ingress paths configuration.

`pathType: Exact` is enough for the GraphQL single endpoint. And I confirmed this works with ALB.
https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types

Screen Shot 2022-07-01 at 20 10 48

(If it is set to `Prefix`, the ALB path becomes `/graphql OR /graphql/*`, but the second is not needed.)

image

### DIFF
--- a/charts/scalardb-graphql/README.md
+++ b/charts/scalardb-graphql/README.md
@@ -17,7 +17,7 @@ Current chart version is `1.0.0`
 | ingress.annotations | object | `{"alb.ingress.kubernetes.io/healthcheck-path":"/graphql?query=%7B__typename%7D","alb.ingress.kubernetes.io/scheme":"internal","alb.ingress.kubernetes.io/target-group-attributes":"stickiness.enabled=true,stickiness.lb_cookie.duration_seconds=60","alb.ingress.kubernetes.io/target-type":"ip","nginx.ingress.kubernetes.io/affinity":"cookie","nginx.ingress.kubernetes.io/session-cookie-hash":"sha1","nginx.ingress.kubernetes.io/session-cookie-max-age":"300","nginx.ingress.kubernetes.io/session-cookie-name":"INGRESSCOOKIE","nginx.ingress.kubernetes.io/session-cookie-path":"/"}` | The class-specific annotations for the ingress resource. |
 | ingress.className | string | `""` | The ingress class name. Specify "alb" for AWS Application Load Balancer. |
 | ingress.enabled | bool | `true` | Enable ingress resource. |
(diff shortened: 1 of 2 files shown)


kota2and3kan (commented):
Thank you for the update!

The new configuration is looks good to me.
However, if we update the `values.yaml`, we need to update `README.md` too.

I think the following test failed since there is a difference between the `values.yaml` and `README.md`.
https://github.com/scalar-labs/helm-charts/runs/7149374330?check_suite_focus=true

Could you please update the `README.md`?

---

For your reference, I was just curious and I tried to this new configuration with NGINX Ingress.
As a result, I was able to run the query via NGINX Ingress with this new configuration.

```console
$ kubectl get ingress
NAME                       CLASS   HOSTS   ADDRESS        PORTS   AGE
graphql-scalardb-graphql   nginx   *       192.168.49.2   80      94s
```
```json
$ kubectl get ingress graphql-scalardb-graphql -o jsonpath='{.spec}' | jq .
{
  "ingressClassName": "nginx",
  "rules": [
    {
      "http": {
        "paths": [
          {
            "backend": {
              "service": {
                "name": "graphql-scalardb-graphql",
                "port": {
                  "number": 8080
                }
              }
            },
            "path": "/graphql",
            "pathType": "Exact"
          }
        ]
      }
    }
  ]
}
```
```console
$ curl http://`minikube ip`/graphql \
>   -H 'Content-Type: application/json' \
>   --data '{"query": "query Get @transaction(commit: true) {t0_get(get: {key: {c1: 1, c2: 11}}) {t0 {c3}}}"}'
{"data":{"t0_get":{"t0":{"c3":"aaa"}}},"extensions":{"transaction":{"id":"eafeb8ec-9d60-4086-bb55-829721def382"}}}
```

ymorimo:
@kota2and3kan Thank you for pointing out the README issue! I was careless.
And thank you very much for trying the values with Nginx!

kota2and3kan (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@ymorimo Some CIs fail, but is it OK?

ymorimo:
> @ymorimo Some CIs fail, but is it OK?

I'll check the situation.

ymorimo:
> @ymorimo Some CIs fail, but is it OK?

@feeblefakie Yes, it's expected. Currently, the Helm chart tries to install scalardb-graphql 3.5.0, but that version needs to be fixed to use `dockerize` so that it can be configured with environment variables. As the next version of scalardb-graphql is released, updating the `appVersion` of this Helm chart should fix the CI problem.

brfrn169 (approved):
LGTM! Thank you!

PR #117, opened 2022-07-20
Remove unnecessary capabilities from Scalar DB and Envoy chert

kota2and3kan:
This PR removes unnecessary Linux capability `NET_BIND_SERVICE` from Scalar DB and Envoy chart.
`NET_BIND_SERVICE` allows binding a socket to internet domain privileged ports (port numbers less than 1024).

https://man7.org/linux/man-pages/man7/capabilities.7.html
https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

However, Scalar DB and Envoy does not bind any ports less than 1024.
So, I remove this unnecessary capability to make them more secure.

Please take a look!


supl (approved):
LGTM.
Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #118, opened 2022-07-20
Add `seccompProfile` field in `securityContext` instead of annotation

kota2and3kan:
This PR adds `securityContext.seccompProfile` in each chart.

---

In Scalar DB and Envoy chart, they are using annotation `seccomp.security.alpha.kubernetes.io/pod` for setting seccomp configuration.
However, this annotation is deprecated and it will be not-functional in next Kubernetes release as follows.

* Message when I deploy current chart
  ```console
  Warning: spec.template.metadata.annotations[seccomp.security.alpha.kubernetes.io/pod]: deprecated since v1.19, non-functional in v1.25+; use the "seccompProfile" field instead
  ```

So, I removed deprecated annotation and added `securityContext.seccompProfile` for setting the same seccomp configuration.

---

In Scalar DL (Ledger and Auditor) chart, there is no configuration related to the seccomp now.
So, I just added `securityContext.seccompProfile` for setting seccomp configuration to make them more secure.

Please take a look!

PR #119, opened 2022-07-20
Disable seccomp checking in CI (kubeaudit)

kota2and3kan:
I updated the security configurations in the following PR.
https://github.com/scalar-labs/helm-charts/pull/118

Now,  in the Scalar Helm Charts, we use `securityContext.seccompProfile` instead of `seccomp.security.alpha.kubernetes.io/pod` annotation. This is because this annotation is deprecated now and it will be removed in the next kubernetes release.
However, the `kubeaudit` returns error as a result of `seccomp` checking since the `kubeaudit` does not understand `PodSecurityContext.seccompProfile`.
This is a known issue in the `kubeaudit`.

Please see the following comment too.
https://github.com/scalar-labs/helm-charts/pull/118#issuecomment-1189894201

So, I updated the configuration of `kubeaudit` to disable `seccomp` checking.
We can enable it again after the known issue of `kubeaudit` will be fixed.

Please take a look!

### DIFF
--- a/.github/kube-audit.yaml
+++ b/.github/kube-audit.yaml
@@ -11,7 +11,7 @@ enabledAuditors:
   privesc: true
   privileged: true
   rootfs: false
-  seccomp: true
+  seccomp: false
 auditors:
   capabilities:
     allowAddList: ['NET_BIND_SERVICE']


feeblefakie (approved):
LGTM! Thank you!

PR #122, opened 2022-07-21
Remove unnecessary envoy default setting from each chart

kota2and3kan:
This PR removes some envoy default values from Scalar DB/DL chart.

Regarding envoy values, the same values are set as default values in Envoy chart and Scalar DB/DL chart.
In other words, there are some redundancy default values between Envoy chart and Scalar DB/DL chart.

Now, even if we update the default values of Envoy chart, it is not applied when we deploy Scalar DB/DL since Scalar DB/DL chart override that default values.
So, we need to update both Envoy and Scalar DB/DL chart if we want to update default values of Envoy chart.

I removed the above redundancy default values.
Also, I remains some values that we need to override to make them work properly.
I will add comment in each remaining values in Scalar DB/DL chart.

Please take a look!

### DIFF
--- a/charts/scalardb/README.md
+++ b/charts/scalardb/README.md
@@ -13,36 +13,15 @@ Current chart version is `2.3.0`
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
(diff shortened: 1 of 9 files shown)


feeblefakie (approved):
Nice refactoring! LGTM, Thank you!

PR #132, opened 2022-08-19
Update scalardb-graphql version to 3.6.0

ymorimo:
This PR updates the version of Scalar DB GraphQL to 3.6.0.

The following changes have also been made in this PR to make the probes work well and for the CI to pass.

1. liveness/readiness probes

    `initialDelaySeconds` was added to liveness/readiness probes to avoid restarting pods in the initial startup.
    The probe method was also changed from `httpGet` to `tcpSocket` since Scalar DB GraphQL 3.6.0 always starts (or continues) a transaction when it receives a GraphQL query and it triggers database access. The liveness/readiness probe should not affect the database.

1. schema-loading job in CI

    In order for the `ct install` command to install the scalardb-graphql chart in CI, it is necessary to load a database schema into the Cassandra database since Scalar DB GraphQL requires at least one table to generate a GraphQL schema. (The GraphQL spec doesn't allow leaving the `Query` root type empty. This is why Scalar DB GraphQL can't start without tables.)
    This PR adds a simple schema loading job.

### DIFF
--- a/.github/workflows/helm_charts_scalar.yml
+++ b/.github/workflows/helm_charts_scalar.yml
@@ -178,6 +178,7 @@ jobs:
           kubectl create -f .github/cassandra.yaml
           echo "Wait until cassandra is running" && sleep 100
           kubectl create -f .github/schema-loading.yaml
+          kubectl wait --for=condition=complete --timeout=60s job/schema-loading
           kubectl get pods,svc,endpoints,nodes -o wide
 
       - name: Run chart-testing (install)
--- a/.github/workflows/helm_charts_scalar.yml
+++ b/.github/workflows/helm_charts_scalar.yml
@@ -177,6 +177,7 @@ jobs:
           kubectl create secret generic auditor-keys --from-file=certificate=.github/auditor-cert.pem --from-file=private-key=.github/auditor-key.pem
           kubectl create -f .github/cassandra.yaml
           echo "Wait until cassandra is running" && sleep 100
+          kubectl create -f .github/schema-loading.yaml
           kubectl get pods,svc,endpoints,nodes -o wide
 
       - name: Run chart-testing (install)
--- a/charts/scalardb-graphql/ci/ct-values.yaml
+++ b/charts/scalardb-graphql/ci/ct-values.yaml
@@ -1,0 +1,2 @@
+scalarDbGraphQlConfiguration:
+  namespaces: ct
--- a/.github/schema-loading.yaml
+++ b/.github/schema-loading.yaml
@@ -1,0 +1,58 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: schema-loading-config
+data:
+  database.properties: |-
+    scalar.db.contact_points=cassandra
+    scalar.db.contact_port=9042
+    scalar.db.username=cassandra
+    scalar.db.password=cassandra
+    scalar.db.storage=cassandra
+  schema.json: |-
+    {
+      "ct.t1": {
+        "transaction": true,
+        "partition-key": [
+          "c1"
+        ],
+        "clustering-key": [
+          "c2"
+        ],
+        "columns": {
+          "c1": "INT",
+          "c2": "INT",
+          "c3": "INT"
+        }
+      }
+    }
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: schema-loading
+spec:
+  template:
+    metadata:
+      labels:
+        app: schema-loading
+    spec:
+      containers:
+      - name: schema-loading
+        image: ghcr.io/scalar-labs/scalardb-schema-loader:3.5.2
+        imagePullPolicy: IfNotPresent
+        args:
+        - "-c"
+        - "/config/database.properties"
+        - "-f"
+        - "/config/schema.json"
+        - "--coordinator"
+        volumeMounts:
+        - name: schema-loading-config-volume
+          mountPath: /config
+      volumes:
+      - name: schema-loading-config-volume
+        configMap:
+          name: schema-loading-config
+      restartPolicy: Never
+  backoffLimit: 0
--- a/charts/scalardb-graphql/Chart.yaml
+++ b/charts/scalardb-graphql/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: scalardb-graphql
 description: Scalar DB GraphQL server
 type: application
-version: 1.0.0
-appVersion: 3.5.0
+version: 1.1.0
+appVersion: 3.6.0
 icon: https://scalar-labs.com/wp-content/themes/scalar/assets/img/logo_scalar.svg
 keywords:
 - scalardb
--- a/charts/scalardb-graphql/README.md
+++ b/charts/scalardb-graphql/README.md
@@ -1,7 +1,7 @@
 # scalardb-graphql
 
 Scalar DB GraphQL server
-Current chart version is `1.0.0`
+Current chart version is `1.1.0`
 
 ## Values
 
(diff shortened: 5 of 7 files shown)


charts/scalardb-graphql/ci/ct-values.yaml
```diff
@@ -0,0 +1,2 @@
+scalarDbGraphQlConfiguration:
+  namespaces: ct
```
ymorimo, line 2: `namespaces` is a required config item for the scalardb-graphql chart.

chart-testing uses the glob pattern `<chart root>/ci/*-values.yaml` to find custom values files.

https://github.com/helm/chart-testing/blob/main/doc/ct_install.md#ct-install

> Charts may have multiple custom values files matching the glob pattern '*-values.yaml' in a directory named 'ci' in the root of the chart's directory. The chart is installed and tested for each of these files. If no custom values file is present, the chart is installed and tested with defaults.

superbrothers:
> 1. liveness/readiness probes
> initialDelaySeconds was added to liveness/readiness probes to avoid restarting pods in the initial startup.

For the above, you can consider using the startup probe, which has been available since Kubernetes 1.18. See the document and entry below for details.

- [Configure Liveness, Readiness and Startup Probes \| Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)
- [Kubernetes 1\.18の新機能を学び、使ってみよう \| Think IT（シンクイット）](https://thinkit.co.jp/article/17500)

.github/workflows/helm_charts_scalar.yml
```diff
@@ -177,6 +177,7 @@ jobs:
           kubectl create secret generic auditor-keys --from-file=certificate=.github/auditor-cert.pem --from-file=private-key=.github/auditor-key.pem
           kubectl create -f .github/cassandra.yaml
           echo "Wait until cassandra is running" && sleep 100
+          kubectl create -f .github/schema-loading.yaml
```
superbrothers, line 180: I recommend that waiting for the job execution to complete makes CI more stable.

```
kubectl wait --for=condition=complete --timeout=60s job/schema-loading
```

superbrothers: It is also better to wait for Pods to be ready for Cassandra as well, as CI is expected to be more stable and finish sooner (no need to wait for 100s).

kota2and3kan: @superbrothers  
Thank you for your comment!
I agree with waiting for Cassandra in another way.

@ymorimo  

I think the `kubectl wait` command is not appropriate as a way to wait for Cassandra to start.
This is because Cassandra is not ready for accepting requests in the container even if the pod (container) status is `Running`.
So, I usually use the following command to wait for Cassandra to accept the requests, in my local environment testing.
```console
echo -n "Wait for Cassandra* "
until kubectl exec cassandra-0 -- cqlsh 2> /dev/null; do  sleep 0.5; echo -n "#"; done
```
This command checks whether Cassandra is ready to accept a request or not using `cqlsh`.
I think we can use the same command in this CI for waiting for Cassandra to start.

.github/workflows/helm_charts_scalar.yml
```diff
@@ -177,6 +177,7 @@ jobs:
           kubectl create secret generic auditor-keys --from-file=certificate=.github/auditor-cert.pem --from-file=private-key=.github/auditor-key.pem
           kubectl create -f .github/cassandra.yaml
           echo "Wait until cassandra is running" && sleep 100
```
kota2and3kan, line 179:
```suggestion
          echo "Wait until cassandra is running" && sleep 100
          echo -n "Wait for Cassandra "
          until kubectl exec cassandra-0 -- cqlsh 2> /dev/null; do  sleep 0.5; echo -n "#"; done
```
As I mentioned in another comment, I think we can use this command for waiting for Cassandra to start.

kota2and3kan (commented):
Thank you for the updates!
Overall looks good to me other than the comments of Suda-san.

@superbrothers  
Thank you for the information about the startup probe.
I didn't know that, but it looks good to me.
I agree with using the startup probe instead of `initialDelaySeconds` for waiting for the startup of Scalar DB GraphQL Server in the pod.

@ymorimo  
I will try to test the startup probe in my local environment.
After that, let's discuss this feature!

kota2and3kan:
I tried to use `startupProbe` as follows.
```diff
$ git diff
diff --git a/charts/scalardb-graphql/templates/scalardb-graphql/deployment.yaml b/charts/scalardb-graphql/templates/scalardb-graphql/deployment.yaml
index 6627208..ec615a4 100644
--- a/charts/scalardb-graphql/templates/scalardb-graphql/deployment.yaml
+++ b/charts/scalardb-graphql/templates/scalardb-graphql/deployment.yaml
@@ -82,14 +82,17 @@ spec:
               value: "{{ .Values.scalarDbGraphQlConfiguration.logLevel }}"
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          startupProbe:
+            tcpSocket:
+              port: 8080
+            failureThreshold: 30
+            periodSeconds: 10
           livenessProbe:
             tcpSocket:
               port: 8080
-            initialDelaySeconds: 60
           readinessProbe:
             tcpSocket:
               port: 8080
-            initialDelaySeconds: 60
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
```

As a result, Scalar DB GraphQL pods are started without restarting by liveness probe.
```console
$ kubectl get pod
NAME                                READY   STATUS    RESTARTS   AGE
scalardb-graphql-6d8c7dc4c8-5jrl4   1/1     Running   0          15m
scalardb-graphql-6d8c7dc4c8-9w9w5   1/1     Running   0          15m
scalardb-graphql-6d8c7dc4c8-mqjw2   1/1     Running   0          15m
```

Also, the startup probe failed several times as expected.
```console
$ kubectl describe pod scalardb-graphql-6d8c7dc4c8-5jrl4

(snip)

Events:
  Type     Reason     Age                From               Message
  ----     ------     ----               ----               -------
  Normal   Scheduled  15m                default-scheduler  Successfully assigned default/scalardb-graphql-6d8c7dc4c8-5jrl4 to minikube
  Normal   Pulled     15m                kubelet            Container image "ghcr.io/scalar-labs/scalardb-graphql:3.6.0" already present on machine
  Normal   Created    15m                kubelet            Created container scalardb-graphql
  Normal   Started    15m                kubelet            Started container scalardb-graphql
  Warning  Unhealthy  14m (x2 over 15m)  kubelet            Startup probe failed: dial tcp 172.17.0.14:8080: connect: connection refused
```

The `startupProbe` seems good to me!

ymorimo:
@kota2and3kan Thank you for testing it! I did the same kind of check and pushed the change.

ymorimo:
@kota2and3kan As for the wait for the Cassandra process, it is not just related to the GraphQL Helm chart, so it might be better to apply the changes in a separate PR.

.github/workflows/helm_charts_scalar.yml
superbrothers: It looks like you can use `kubectl wait` command by setting up the readiness probe for cassandra Pod.

```
diff --git a/.github/cassandra.yaml b/.github/cassandra.yaml
index e0a8a67..fc0e677 100644
--- a/.github/cassandra.yaml
+++ b/.github/cassandra.yaml
@@ -21,6 +21,9 @@ spec:
           env:
             - name: CASSANDRA_SEEDS
               value: "cassandra-0.cassandra-headless.default.svc.cluster.local"
+          readinessProbe:
+            exec:
+              command: ["cqlsh", "-e", "SELECT now() FROM system.local;"]
 ---
 apiVersion: v1
 kind: Service
```

```
kubectl wait --for=condition=Ready pod/cassandra-0
```

.github/workflows/helm_charts_scalar.yml
superbrothers: It looks like you can use `kubectl wait` command by setting up the readiness probe for cassandra Pod.

```
diff --git a/.github/cassandra.yaml b/.github/cassandra.yaml
index e0a8a67..fc0e677 100644
--- a/.github/cassandra.yaml
+++ b/.github/cassandra.yaml
@@ -21,6 +21,9 @@ spec:
           env:
             - name: CASSANDRA_SEEDS
               value: "cassandra-0.cassandra-headless.default.svc.cluster.local"
+          readinessProbe:
+            exec:
+              command: ["cqlsh", "-e", "SELECT now() FROM system.local;"]
 ---
 apiVersion: v1
 kind: Service
```

```
kubectl wait --for=condition=Ready pod/cassandra-0
```

However, this change should be implemented in a separate PR.

kota2and3kan:
@ymorimo  
> As for the wait for the Cassandra process, it is not just related to the GraphQL Helm chart, so it might be better to apply the changes in a separate PR.

Ah, I see. That's right.
I agree with updating it in another PR.

.github/workflows/helm_charts_scalar.yml
kota2and3kan: @superbrothers  
We will update it in another PR.

superbrothers (approved):

kota2and3kan (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #126, opened 2022-08-04
Update chart version to keep main branch latest (Ledger)

kota2and3kan:
A new minor version of Scalar DL Ledger Helm Charts has been released.
This PR updates version of Scalar DL Ledger chart to keep main branch latest.
(This release flow will be fixed in the future.)

This PR apply the same update as the following commit.
https://github.com/scalar-labs/helm-charts/commit/09f98fabf65bed61bcb5e6273af97164b1577fc7

Please take a look!

PR #144, opened 2022-10-05
Update chart version to keep main branch latest (Scalar DB)

kota2and3kan:
A new minor version of Scalar DB Helm Charts has been released.
This PR updates version of Scalar DB chart to keep main branch latest.
(This release flow will be fixed in the future.)

This PR apply the same update as the following commit.
https://github.com/scalar-labs/helm-charts/commit/3c265f7aff8ea7319e9cdf819c3c3f935b2cb14d

Please take a look!

### DIFF
--- a/charts/scalardb/Chart.lock
+++ b/charts/scalardb/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: envoy
   repository: https://scalar-labs.github.io/helm-charts
-  version: 2.1.0
-digest: sha256:21cca64d71b44b1da064db4c42be1a19dd230415494752b58698aaa8bd9a5d77
-generated: "2022-07-11T14:16:08.15899893+09:00"
+  version: 2.2.0
+digest: sha256:2400613e162d0acb575e173d8623a67281e131d0e00b0da7925177e793c13643
+generated: "2022-09-14T12:34:38.629511121+09:00"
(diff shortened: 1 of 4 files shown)


feeblefakie (approved):
LGTM! Than you!

feeblefakie:
@kota2and3kan 
BTW, why [Chart.lock](https://github.com/scalar-labs/helm-charts/pull/144/files#diff-770ecb29b15bfda989bc80400319ae3a65e23f3dfd9a0a53cc5328545eb82521) was updated in Scalar DB but it was not updated in Scalar DL?

kota2and3kan:
@feeblefakie  
The update of `Chart.lock` is related to the dependency of Envoy.
In other words, if we update the minor version of the Scalar Envoy Chart, we need to update `Chart.lock` in the next minor release of the Scalar DB and Scalar DL chart.

At this time, a minor version up of Scalar DB Helm Chart (v2.3 -> v2.4) includes the update of the Scalar Envoy chart (v2.1 -> v2.2).
So, we need to update the dependency configuration about Envoy in the `Chart.lock` of the Scalar DB chart.

And, the Scalar DL chart has already included the Envoy chart minor update (v2.1 -> v2.2) in the previous minor version up (v4.2 -> v4.3).
You can see it in the following PR.
https://github.com/scalar-labs/helm-charts/pull/126/files

So, in the update of this time (v4.3 -> v4.4), we don't need to update the dependency of Envoy in the `Chart.lock` of Scalar DL chart.
This is why there is a difference between the Scalar DB chart update and the Scalar DL chart update in this time.

PR #145, opened 2022-10-05
Increase timeout configuration for Cassandra in the CI

kota2and3kan:
In the following PR, the CI failed several times with the timeout error in the step of waiting for Cassandra starts.
https://github.com/scalar-labs/helm-charts/pull/142

I re-run it several times and finally it succeeded, but I think it is better to increase the timeout configuration.
So, I updated the timeout configuration from `60s` to `120s`.

Please take a look!


kota2and3kan:
In the CI of this PR, it succeeded the first time. The timeout error didn't occur.

feeblefakie (approved):
LGTM! Thank you!

PR #155, opened 2022-12-08
Tmp test all charts

kota2and3kan:
This PR is used for temporary testing. Don't merge this PR to the main branch.

PR #158, opened 2022-12-09
Add supported k8s version in the README

kota2and3kan:
This PR adds the description of the supported Kubernetes version.
It is decided according to the test result in the following PR.
https://github.com/scalar-labs/helm-charts/pull/155

Please take a look!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,7 @@ This directory contains the following helm charts.
 
 ## Supported Kubernetes versions
 
-* Kubernetes 1.25
-* Kubernetes 1.24
-* Kubernetes 1.23
-* Kubernetes 1.22
-* Kubernetes 1.21
+* 1.25.x, 1.24.x, 1.23.x, 1.22.x, 1.21.x
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -13,8 +13,15 @@ This directory contains the following helm charts.
 
 ## Prerequisites
 
-* Kubernetes 1.19+
 * Helm 3.5+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,9 +13,16 @@ This directory contains the following helm charts.
 
 ## Prerequisites
 
-* Kubernetes 1.19+
 * Helm 3.5+
 
+## Supported Kubernetes version
```
feeblefakie, line 18:
```suggestion
## Supported Kubernetes versions
```

README.md
```diff
@@ -13,9 +13,16 @@ This directory contains the following helm charts.
...
+
+* Kubernetes 1.25
```
feeblefakie, line 20: Better to be simplified like this?
```
* 1.[21-25].x
```

feeblefakie (commented):
Thank you! Left one minor suggestion.

README.md
kota2and3kan: Thank you for your suggestion!

It is simple and we can maintain it easily. So, I think it is better from perspective of developer/maintainer (i.e., us).
However, I think there is a possibility that the regular expressions is a bit confusing.

The `list` is easy to understand with a quick glance.
So, from the perspective of users, I think it is better to describe supported version as a list.

What do you think?

kota2and3kan:
I updated the description based on the result of the discussion.
Please take a look again!

feeblefakie (approved):
LGTM! Thank you!

PR #168, opened 2023-01-19
Add sleep command before kubectl wait in the CI

kota2and3kan:
This PR adds the `sleep 1` command before `kubectl wait` in the CI.

The CI sometimes fails with the following error.

```shell
$ kubectl wait --for=condition=Ready --timeout=120s pod/cassandra-0
Error from server (NotFound): pods "cassandra-0" not found
```

For example, you can see it in the following action.
https://github.com/scalar-labs/helm-charts/actions/runs/3945941786/jobs/6753281153

I guess, there is a time lag between the `StatefulSet` creation and the `Pod` creation by the StatefulSet.

Also, I investigated the `kubectl wait` command. However, it seems that there is no way to handle (re-try) the `not found` error.
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#wait

So, I added the `sleep` command before the `kubectl wait` command for waiting for the pod creation.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #174, opened 2023-02-15
Fix typo of Grafana custom values

kota2and3kan:
This PR fixes a typo in the getting started with Scalar Manager.
Regarding the key name, `grafana.ini` is correct.

https://github.com/scalar-labs/helm-charts/blob/main/docs/conf/scalar-prometheus-custom-values.yaml#L35
https://github.com/grafana/helm-charts/blob/grafana-6.50.7/charts/grafana/values.yaml#L692


supl (approved):
Thank you.

I didn't notice this, and I checked my local custom values and that was `grafana.ini` 😄

feeblefakie (approved):
LGTM! Thank you!

PR #182, opened 2023-03-17
Allow specifying any serviceaccounts in Ledger chart

kota2and3kan:
This PR adds a new feature that can set an arbitrary Service Account to the ScalarDL Ledger pods.

For example, if we use a pay-as-you-go container in an EKS cluster, we need to create Service Account to grant permission to run the `AWS Marketplace Metering Service API` and specify it to `ledger.serviceAccount.serviceAccountName` in this chart. It uses the [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) feature of AWS.

Please take a look!

---

For your reference, you can create the Service Account to allow running `AWS Marketplace Metering Service API` as follows.
```console
eksctl create iamserviceaccount \
    --name <SERVICE_ACCOUNT_NAME> \
    --namespace <NAMESPACE> \
    --cluster <ENTER_YOUR_CLUSTER_NAME_HERE> \
    --attach-policy-arn arn:aws:iam::aws:policy/AWSMarketplaceMeteringFullAccess \
    --attach-policy-arn arn:aws:iam::aws:policy/AWSMarketplaceMeteringRegisterUsage \
    --attach-policy-arn arn:aws:iam::aws:policy/service-role/AWSLicenseManagerConsumptionPolicy \
    --approve \
    --override-existing-serviceaccounts
```

And, you can specify the above Service Account name as follows in this chart.
```yaml
ledger:
  serviceAccount:
    serviceAccountName: "<SERVICE_ACCOUNT_NAME>"
    automountServiceAccountToken: true
```

### DIFF
--- a/charts/scalardl/values.yaml
+++ b/charts/scalardl/values.yaml
@@ -227,3 +227,7 @@ ledger:
 
   # -- Name of existing secret to use for storing database username and password
   existingSecret: ""
+
+  serviceAccount:
+    serviceAccountName: ""
+    automountServiceAccountToken: false
--- a/charts/scalardl/values.yaml
+++ b/charts/scalardl/values.yaml
@@ -231,5 +231,5 @@ ledger:
   serviceAccount:
     # -- Name of the existing service account resource
     serviceAccountName: ""
-    # -- Specify to mount secret or not
+    # -- Specify to mount a service account token or not
     automountServiceAccountToken: false
--- a/charts/scalardl/values.yaml
+++ b/charts/scalardl/values.yaml
@@ -229,5 +229,7 @@ ledger:
   existingSecret: ""
 
   serviceAccount:
+    # -- Name of the existing service account resource
     serviceAccountName: ""
+    # -- Specify to mount secret or not
     automountServiceAccountToken: false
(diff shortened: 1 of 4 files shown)


supl (approved):
LGTM.
Thank you 🙇

charts/scalardl/values.yaml
```diff
@@ -227,3 +227,9 @@ ledger:
 
   # -- Name of existing secret to use for storing database username and password
   existingSecret: ""
+
+  serviceAccount:
+    # -- Name of the existing service account resource
+    serviceAccountName: ""
+    # -- Specify to mount secret or not
```
feeblefakie, line 234: Should we say `service account` for consistency?
```suggestion
    # -- Specify to mount a service account token or not
```

feeblefakie (approved):
LGTM! Thank you!
Left one minor comment, so PTAL before merge!

charts/scalardl/values.yaml
kota2and3kan: Thank you for your suggestion!
I think it is better than the current description.
So, I will apply it. Thank you!

PR #184, opened 2023-03-17
Allow specifying any serviceaccounts in Scalar Manager chart

kota2and3kan:
This PR adds a new feature that can set an arbitrary Service Account to the Scalar Manager pods.

For example, if we use a pay-as-you-go container in an EKS cluster, we need to create Service Account to grant permission to run the `AWS Marketplace Metering Service API` and specify it to `serviceAccount.serviceAccountName` in this chart. It uses the [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) feature of AWS.

Also, the Scalar Manager pod already uses Service Account to allow running the Kubernetes API from inside of Scalar Manager pod.
We cannot set two Service Accounts to one pod, so I adjusted `Role` and `Rolebinding` resource.
If we don't specify the existing Service Account, this chart generates Service Account automatically and grants some permissions to it.
If we specify the existing Service Account, this chart grants some permissions to the specified Service Account.

Please take a look!

---

For your reference, you can create the Service Account to allow running `AWS Marketplace Metering Service API` as follows.
```console
eksctl create iamserviceaccount \
    --name <SERVICE_ACCOUNT_NAME> \
    --namespace <NAMESPACE> \
    --cluster <ENTER_YOUR_CLUSTER_NAME_HERE> \
    --attach-policy-arn arn:aws:iam::aws:policy/AWSMarketplaceMeteringFullAccess \
    --attach-policy-arn arn:aws:iam::aws:policy/AWSMarketplaceMeteringRegisterUsage \
    --attach-policy-arn arn:aws:iam::aws:policy/service-role/AWSLicenseManagerConsumptionPolicy \
    --approve \
    --override-existing-serviceaccounts
```

And, you can specify the above Service Account name as follows in this chart.
```yaml
serviceAccount:
  serviceAccountName: "<SERVICE_ACCOUNT_NAME>"
  automountServiceAccountToken: true
```

### DIFF
--- a/charts/scalar-manager/values.yaml
+++ b/charts/scalar-manager/values.yaml
@@ -20,10 +20,10 @@ nameOverride: ""
 fullnameOverride: ""
 
 serviceAccount:
-  # serviceAccount.create -- Create a service account
-  create: true
-  # serviceAccount.name -- Specify the name for the service account
-  name: scalar-manager
+  # serviceAccount.serviceAccountName -- Name of the existing service account resource
+  serviceAccountName: ""
+  # serviceAccount.automountServiceAccountToken -- Specify to mount secret or not
+  automountServiceAccountToken: true
 
 service:
   # service.type -- The service type
--- a/charts/scalar-manager/values.yaml
+++ b/charts/scalar-manager/values.yaml
@@ -22,7 +22,7 @@ fullnameOverride: ""
 serviceAccount:
   # serviceAccount.serviceAccountName -- Name of the existing service account resource
   serviceAccountName: ""
-  # serviceAccount.automountServiceAccountToken -- Specify to mount secret or not
+  # serviceAccount.automountServiceAccountToken -- Specify to mount a service account token or not
   automountServiceAccountToken: true
 
 service:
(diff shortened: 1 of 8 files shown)


supl (approved):
LGTM.
Thank you 🙇

charts/scalar-manager/values.yaml
```diff
@@ -20,10 +20,10 @@ nameOverride: ""
 fullnameOverride: ""
 
 serviceAccount:
-  # serviceAccount.create -- Create a service account
-  create: true
-  # serviceAccount.name -- Specify the name for the service account
-  name: scalar-manager
+  # serviceAccount.serviceAccountName -- Name of the existing service account resource
+  serviceAccountName: ""
+  # serviceAccount.automountServiceAccountToken -- Specify to mount secret or not
```
feeblefakie, line 25: Ditto.

feeblefakie (approved):
LGTM! Thank you!
Left the same comment as the other related PRs.

charts/scalar-manager/values.yaml
kota2and3kan: Thank you for your review!
I will update this description.

PR #192, opened 2023-04-11
Remove unnecessary environment variables from ScalarDB Cluster chart

kota2and3kan:
This PR removes unnecessary environment variables from the ScalarDB Cluster chart.

Note: This PR is related to the following PR of the ScalarDB Cluster.
https://github.com/scalar-labs/scalardb-cluster/pull/66

ScalarDB Cluster uses the following two environment variables for running Kubernetes API (specify the Kubernetes API's endpoint) in its internal.

* KUBERNETES_SERVICE_HOST
* KUBERNETES_SERVICE_PORT

And, in the current chart, we set these environment variables as fixed values.

However, these two environment variables are set automatically by Kubernetes (strictly, set by Kubelet that one of the components of Kubernetes) in all pods.

```shell
$ kubectl run test-pod --image=ubuntu:20.04 -- sleep inf
pod/test-pod created
```
```shell
$ kubectl get pod
NAME       READY   STATUS    RESTARTS   AGE
test-pod   1/1     Running   0          34s
```
```shell
$ kubectl exec -it test-pod -- env | grep -i kubernetes_service
KUBERNETES_SERVICE_HOST=10.96.0.1
KUBERNETES_SERVICE_PORT=443
KUBERNETES_SERVICE_PORT_HTTPS=443
```

Note: The official document [Connecting Applications with Services](https://kubernetes.io/docs/tutorials/services/connect-applications-service/#environment-variables) describes this behavior.

As you can see, Kubelet set the appropriate values, so we don't need to set them manually.

Also, there are cons when we set fixed values. If we set fixed values and if the Kubernetes API's endpoint is changed by some Kubernetes' configuration or version up, ScalarDB Cluster cannot run the Kubernetes API and we will face some errors.

So, I think it is better to use the automatically set (default) environment variables to get the Kubernetes API endpoint information.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@kota2and3kan Sorry, can you fix the conflict?

kota2and3kan:
@feeblefakie  
Thank you for your review!
I will fix the conflict. And after that, I will merge this PR.

PR #194, opened 2023-04-11
Update CI for ScalarDB Cluster

kota2and3kan:
This PR updates the CI for ScalarDB Cluster.

The `ct` and `kubeval` commands get all chart names (directories) automatically in the current CI. So, we don't need to update their workflow.

However, we specify the list of chart (directory) names for the `kubeaudit` testing. So, I added the chart name `scalardb-cluster` to the list.

Please take a look!


feeblefakie (approved):
LGTM! Thank you!

PR #197, opened 2023-04-12
Fix comment in scalardb-cluster-node.properties file

kota2and3kan:
This PR fixes the comment about the default value of `scalar.db.graphql.enabled` in the `values.yaml` file of the ScalarDB Cluster chart.

We set `scalar.db.graphql.enabled` to `false` by default in the following commit of ScalarDB Cluster.
https://github.com/scalar-labs/scalardb-cluster/commit/e35e616ae79cb082314d6c88ef662138dd6001f4

So, I fixed the comment based on the above default value.

Please take a look!

### DIFF
--- a/charts/scalardb-cluster/values.yaml
+++ b/charts/scalardb-cluster/values.yaml
@@ -138,7 +138,7 @@ scalardbCluster:
     # For GraphQL configurations
     #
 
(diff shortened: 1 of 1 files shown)


feeblefakie (approved):
LGTM! Thank you!

PR #208, opened 2023-04-28
Add further reading part in getting started guide

kota2and3kan:
This PR adds the links to some guides for monitoring/logging to the getting started guide of ScalarDB/ScalarDL.

From the perspective of our documentation site, we want to list only necessary documents in the left pane of the documentation site. This is because, listing all documents is a bit confusing since there are a lot of product's documents.

To achieve this, we decided to list only three documents as root documents of Scalar Helm Chart.
* Getting Started with Helm Charts (ScalarDB Server)
* Getting Started with Helm Charts (ScalarDL Ledger / Ledger only)
* Getting Started with Helm Charts (ScalarDL Ledger and Auditor / Auditor mode)

However, there is no information about monitoring/logging in the above documents. So, I added links to those documents to lead users to the monitoring/logging documents.

Please take a look!

### DIFF
--- a/docs/getting-started-scalardb.md
+++ b/docs/getting-started-scalardb.md
@@ -354,3 +354,11 @@ After completing the ScalarDB Server tests on the Kubernetes cluster, remove all
    ```
    kubectl delete pod scalardb-client --force --grace-period 0
    ```
+
+## Further reading
+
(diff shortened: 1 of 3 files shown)


feeblefakie (approved):
LGTM! Thank you!

josh-wong (approved):
LGTM! Thank you!🙏

PR #220, opened 2023-05-15
[envoy] Add readiness and startup probe

kota2and3kan:
This PR adds `Readiness Probe` and `Startup Probe` in the Envoy chart.

At this time (no readiness probe), there is a possibility that the CI succeeded even if there are some problems. The `ct` command checks whether the pod status is `Ready`, and if the status is `Ready` it returns `success` in the CI. However, in the current charts, there is no `Readiness Probe` configuration. So, the pod will be the `Ready` status temporarily when the container starts even if the Envoy process will failed in the container.

As this above, There is a possibility that the `ct` command detects this temporary `Ready` state and returns `success`. To fix this issue, I added `Readiness Probe`. After this update, the pod status will be `Ready` only after Envoy returns  `Ready` in its health check endpoint. So, the `ct` command can check the status of the Envoy process in the container properly in the CI.

Also, I added `Startup Probe` instead of `initialDelaySeconds` of `Readiness/Liveness Probe`. This is a better way in the Kubernetes environment.

On the other hand, from the perspective of users, this update can increase the reliability of the deployment of Scalar products since Kubernetes runs the health check better way.

Please take a look!

### DIFF
--- a/charts/envoy/templates/deployment.yaml
+++ b/charts/envoy/templates/deployment.yaml
@@ -59,14 +59,6 @@ spec:
               port: 9001
             failureThreshold: 60
             periodSeconds: 5
-          readinessProbe:
-            httpGet:
-              path: /ready
-              port: 9001
-            failureThreshold: 3
-            periodSeconds: 10
-            successThreshold: 1
-            timeoutSeconds: 1
           livenessProbe:
             httpGet:
               path: /ready
--- a/charts/envoy/templates/deployment.yaml
+++ b/charts/envoy/templates/deployment.yaml
@@ -53,12 +53,25 @@ spec:
             value: {{ include "envoy.fullname" . }}-headless
           - name: service_listeners
             value: "{{ .Values.envoyConfiguration.serviceListeners }}"
+          startupProbe:
+            httpGet:
+              path: /ready
+              port: 9001
+            failureThreshold: 60
+            periodSeconds: 5
+          readinessProbe:
+            httpGet:
+              path: /ready
+              port: 9001
+            failureThreshold: 3
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
           livenessProbe:
             httpGet:
               path: /ready
               port: 9001
             failureThreshold: 3
-            initialDelaySeconds: 10
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1


charts/envoy/templates/deployment.yaml
```diff
@@ -53,12 +53,25 @@ spec:
             value: {{ include "envoy.fullname" . }}-headless
           - name: service_listeners
             value: "{{ .Values.envoyConfiguration.serviceListeners }}"
+          startupProbe:
+            httpGet:
+              path: /ready
+              port: 9001
+            failureThreshold: 60
+            periodSeconds: 5
```
kota2and3kan, line 61: I set the timeout value to 300sec (`failureThreshold * periodSeconds = 300sec`). If the pod won't be `Ready` state after 300sec, Kubernetes terminates the pod.

charts/envoy/templates/deployment.yaml
```diff
@@ -53,12 +53,25 @@ spec:
...
+            failureThreshold: 3
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
           livenessProbe:
             httpGet:
               path: /ready
               port: 9001
             failureThreshold: 3
-            initialDelaySeconds: 10
```
kota2and3kan, line 61: We can delete `initialDelaySeconds` since we use `Startup Probe` instead.

charts/envoy/templates/deployment.yaml
```diff
@@ -53,12 +53,25 @@ spec:
             value: {{ include "envoy.fullname" . }}-headless
           - name: service_listeners
             value: "{{ .Values.envoyConfiguration.serviceListeners }}"
+          startupProbe:
+            httpGet:
+              path: /ready
+              port: 9001
+            failureThreshold: 60
+            periodSeconds: 5
+          readinessProbe:
```
feeblefakie, line 62: I think we used to specify the same configuration for `readinessProbe` and `livenessProbe` but removed the `redinessProbe` for some reasons like redundant checking.
So, I'm wondering if putting the `redinessProbe` back really makes things different, or having `startupProbe` is good enough.

BTW, ideally, `readinessProbe` should check the backend services' (ledgers' and auditors') health to identify if a target container/pod is ready to accept traffic , doesn't it?

feeblefakie (commented):
Thank you! Left one question. PTAL!

charts/envoy/templates/deployment.yaml
kota2and3kan: @feeblefakie  
Thank you for your review!

> I think we used to specify the same configuration for `readinessProbe` and `livenessProbe` but removed the `redinessProbe` for some reasons like redundant checking.
> So, I'm wondering if putting the `redinessProbe` back really makes things different, or having `startupProbe` is good enough.

Ah, I see. One of the main purposes of this PR is to return the `Ready` status properly for CI (`ct` command).
I will check if the `startupProbe` (without `readinessProbe`) can return the `Ready` state properly.

> BTW, ideally, `readinessProbe` should check the backend services' (ledgers' and auditors') health to identify if a target container/pod is ready to accept traffic , doesn't it?

Yes, I agree. But, I have some little concerns to check another pod's status by using `readinessProbe`.

From the technical perspective, we can do it by using the `exec` method that can run any commands in the pod for checking another pod's status. But, there is a possibility that it makes some dependencies between two pods, and those dependencies might cause some unexpected issues.

As another concern, I think we need to decide on some rules and implement some logic for checking all pods for the `exec` method. For example:
  * If there is one failed Ledger pod and two ready Ledger pods, should we return `Not Ready` as Envoy state?
  * Or, should we return `Ready` as Envoy state if at least one pod is available?
  * How should we do the above processes (using shell scripts or any tools)?

(If we check the one pod only like RDB, I think it is easy to check its state from another pod. But, Ledger or Auditor has several pods. So, we have to treat them properly in our custom health checking for Ledgers or Auditors.)

I think we need additional development/maintenance costs for these custom health checking. So, if we should use custom health checking, I would like to discuss it as another new task.

Also, at this time, we can detect the `Ready` of the Ledger's or Auditor's state using the following two features.

1. Health check feature of Envoy
   Envoy has a health check feature. So, if one Ledger pod is not healthy, Envoy detects it and doesn't send new requests to the unhealthy pod.
1. Service Discovery by using the `Service` resource
   Envoy uses the `Service` resource of Kubernetes for service (i.e., available pod) discovery and it works as follows.
   1. Ledger returns `SERVING` in the gRPC Health Check.
   1. Kubernetes' `readiness probe` detect `SERVING` and set pod status to `Ready`.
   1. Kubernetes' `Service` resource (strictly controller) detects `the pod status is Ready`.
   1. Kubernetes' `Service` resource adds that ready pod into the list (`Endpoint` resource) as a target of load balancing.
   1. Envoy detects the update of the `Service (Endpoint)` resource (i.e., detects added new ready pod) and adds it to Envoy's own upstream list that includes target pod information.

If the pod state moves to unhealthy (Ledger returns `NOT_SERVING`), Kubernetes removes it from the `Endpoint` resource, and Envoy can detect it and remove the pod from the upstream list.

These steps basically use each component's native feature and we don't need additional costs to implement our own health checking method. So, I think it is enough at this time.

I would like to discuss whether we need additional health checking features or we use the current functions.
What do you think?

kota2and3kan: @feeblefakie  
> I think we used to specify the same configuration for `readinessProbe` and `livenessProbe` but removed the `redinessProbe` for some reasons like redundant checking.
> So, I'm wondering if putting the `redinessProbe` back really makes things different, or having `startupProbe` is good enough.

Thank you for pointing out this. I investigated the behavior of probes. As a result, it is enough to set only `readinessProbe` in our use case. So, I will remove `readinessProbe`.

---

## The details of the investigation result

### Behavior of Kubernetes

The code (`UpdatePodStatus()`) that updates the container status is as follows.
https://github.com/kubernetes/kubernetes/blob/v1.27.2/pkg/kubelet/prober/prober_manager.go#L258

First, it checks the result of the `startupProbe`, and if it is `success` it sets `true` to `pod.status.containerStatuses.started`.
```go
		var started bool
		if c.State.Running == nil {
			started = false
		} else if result, ok := m.startupManager.Get(kubecontainer.ParseContainerID(c.ContainerID)); ok {
			started = result == results.Success
		} else {
			// The check whether there is a probe which hasn't run yet.
			_, exists := m.getWorker(podUID, c.Name, startup)
			started = !exists
		}
		podStatus.ContainerStatuses[i].Started = &started
```

After that (only if the `pod.status.containerStatuses.started = true`), it checks the result of the `readinessProbe`. And, if we don't configure the `readinessProbe`, it always set `true` to `pod.status.containerStatuses.ready`.

```go
		if started {
			var ready bool
			if c.State.Running == nil {
				ready = false
			} else if result, ok := m.readinessManager.Get(kubecontainer.ParseContainerID(c.ContainerID)); ok && result == results.Success {
				ready = true
			} else {
				// The check whether there is a probe which hasn't run yet.
				w, exists := m.getWorker(podUID, c.Name, readiness)
				ready = !exists // no readinessProbe -> always ready
				if exists {
					// Trigger an immediate run of the readinessProbe to update ready state
					select {
					case w.manualTriggerCh <- struct{}{}:
					default: // Non-blocking.
						klog.InfoS("Failed to trigger a manual run", "probe", w.probeType.String())
					}
				}
			}
			podStatus.ContainerStatuses[i].Ready = ready
		}
```

As the above, if we set `startupProbe` and don't set `readinessProbe`, the pod status (whether ready) will be updated based on the result of `startupProbe`. So, we can guarantee the pod status will be `Ready` if and only if after the Scalar product process started properly in the container by using `startupProbe` only without `readinessProbe`.

Also, you can see the explanation of this behavior in the API reference as well.

* `pod.status.containerStatuses.started`
  ```console
  $ kubectl explain pod.status.containerStatuses.started
  KIND:       Pod
  VERSION:    v1
  
  FIELD: started <boolean>
  
  DESCRIPTION:
      Started indicates whether the container has finished its postStart lifecycle
      hook and passed its startup probe. Initialized as false, becomes true after
      startupProbe is considered successful. Resets to false when the container is
      restarted, or if kubelet loses state temporarily. In both cases, startup
      probes will run again. Is always true when no startupProbe is defined and
      container is running and has passed the postStart lifecycle hook. The null
      value must be treated the same as false.
  ```
* `pod.status.containerStatuses.ready`
  ```console
  $ kubectl explain pod.status.containerStatuses.ready
  KIND:       Pod
  VERSION:    v1
  
  FIELD: ready <boolean>
  
  DESCRIPTION:
      Ready specifies whether the container is currently passing its readiness
      check. The value will change as readiness probes keep executing. If no
      readiness probes are specified, this field defaults to true once the
      container is fully started (see Started field).
  
      The value is typically used to determine whether a container is ready to
      accept traffic.
  ```

### Requirements in our (Scalar product's) use case

In our use case, the requirements are the following.

1. We want to guarantee that the pod status will be `Ready` if and only if after the Scalar product process started properly in the container.
1. After the pod started once, we don't need to check the readiness of the pod. This is because it is enough to restart the pod by `livenessProbe` if some failure occurs.

### Conclusion

We can satisfy our requirements by using only `startupProbe`.

kota2and3kan:
@feeblefakie  
I updated this PR (removed `readinessProbe`) based on your feedback and my investigation result. You can see the details of the investigation result in the following comment.

Please take a look again when you have time!

supl (approved):
LGTM.
Thank you!

feeblefakie (approved):
LGTM! Thank you!

superbrothers (commented):
LGTM!

PR #225, opened 2023-05-15
[scalardb-cluster] Add startup probe in ScalarDB Cluster chart

kota2and3kan:
This PR adds `Startup Probe` in the ScalarDB Cluster chart.

Sometimes, ScalarDB Cluster takes a bit long time to start and return `Ready` state (i.e., `SERVING` in gRPC health check). However the `initialDelaySeconds` of `Readiness/Liveness Probe` is fixed value. In other words, if it takes over 60sec to start process, the pod will be killed by Kubernetes and there is a possibility that the `CrashLoopBackOff`.

To avoid this issue, we can use `Startup Probe` instead of `initialDelaySeconds`. This is a better way in the Kubernetes environment to handle health checking when that pod starting process.

We can run the health check before `Readiness/Liveness Probe` starts when the pod starting process by using `Startup Probe`. After only `Startup Probe` succeeded, Kubernetes runs `Readiness/Liveness Probe`. So, we can avoid unexpected pod killing by `Liveness Probe` even if ScalarDB Cluster take a bit long time to starts (I set the timeout 300sec by default).

Please take a look!

### DIFF
--- a/charts/scalardb-cluster/templates/scalardb-cluster/deployment.yaml
+++ b/charts/scalardb-cluster/templates/scalardb-cluster/deployment.yaml
@@ -59,13 +59,19 @@ spec:
           {{- end }}
           resources:
             {{- toYaml .Values.scalardbCluster.resources | nindent 12 }}
+          startupProbe:
+            exec:
+              command:
+                - /usr/local/bin/grpc_health_probe
+                - -addr=:60053
+            failureThreshold: 60
+            periodSeconds: 5
           livenessProbe:
             exec:
               command:
               - /usr/local/bin/grpc_health_probe
               - -addr=:60053
             failureThreshold: 3
-            initialDelaySeconds: 10
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1
@@ -75,7 +81,6 @@ spec:
                 - /usr/local/bin/grpc_health_probe
                 - -addr=:60053
             failureThreshold: 3
-            initialDelaySeconds: 10
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1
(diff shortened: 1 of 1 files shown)


charts/scalardb-cluster/templates/scalardb-cluster/deployment.yaml
```diff
@@ -59,13 +59,19 @@ spec:
           {{- end }}
           resources:
             {{- toYaml .Values.scalardbCluster.resources | nindent 12 }}
+          startupProbe:
+            exec:
+              command:
+                - /usr/local/bin/grpc_health_probe
+                - -addr=:60053
+            failureThreshold: 60
+            periodSeconds: 5
```
kota2and3kan, line 68: I set the timeout value to 300sec (`failureThreshold * periodSeconds = 300sec`). If the pod won't be `Ready` state after 300sec, Kubernetes terminates the pod.

charts/scalardb-cluster/templates/scalardb-cluster/deployment.yaml
```diff
@@ -59,13 +59,19 @@ spec:
...
           livenessProbe:
             exec:
               command:
               - /usr/local/bin/grpc_health_probe
               - -addr=:60053
             failureThreshold: 3
-            initialDelaySeconds: 10
```
kota2and3kan, line 68: We can delete `initialDelaySeconds` since we use `Startup Probe` instead.

brfrn169 (approved):
LGTM! Thank you!

kota2and3kan:
According to the discussion and investigation result in the following comment, I removed `readinessProbe`.

feeblefakie (approved):
LGTM! Thank you!

PR #231, opened 2023-06-07
Add a redirect from the docs site hosted in this repository to a product selection page on the new docs site

josh-wong:
This PR adds files related to Jekyll (static site generator) to add the following functionality:

- Automatically redirect visitors from the previous Scalar Helm Charts docs site (hosted in this repository) to a [product selection page on the new Scalar docs site](https://developers.scalar-labs.com/docs/scalar-helm-charts-product-selection). From there, visitors can choose the Scalar product they are using.
  - Options on the new page include ScalarDB, ScalarDB Community, and ScalarDL.
  - Visitors need to be given an option to choose the product they are using because we do not have a general Scalar Helm Charts docs site. Instead, the Scalar Helm Charts docs are included in the product docs sites.
- Track visits to the previous Scalar Helm Charts docs site (hosted in this repository) so that we can decommission it in the future (after confirming minimal or no impact on visitors).

I have tested the redirect (with different URLs) in my personal repository. In my test, I confirmed the redirect works as expected. Regardless of the page the visitor lands on on the previous docs site, they will be redirected to the product selection page on the new Scalar docs site.


feeblefakie (approved):
LGTM! Thank you!

_config.yml
```diff
@@ -0,0 +1,9 @@
+# Note: This file is used only for redirecting visitors from the old Helm Charts docs site hosted in this repository to a page that gives users the option to go to the ScalarDB, ScalarDB Community, or ScalarDL docs site. This file can be deleted after the old Helm Charts docs site hosted in this repository is no longer deployed.
+google_analytics: "G-Q4TKS77KCP"
+
+defaults:
+  - scope:
+      path: "docs" # Specifies where the docs are located.
```
kota2and3kan, line 6: In this configuration, do we only redirect the URL that includes `docs` in the path?

If the `old Helm Charts docs site` means/includes `https://scalar-labs.github.io/helm-charts/`, I have one concern.

We use `https://scalar-labs.github.io/helm-charts/` as a `Helm Repository`. In other words, users specify this URL to get some artifact of Scalar Helm Chart from this URL as follows.

```
$ helm repo add scalar-labs https://scalar-labs.github.io/helm-charts
```

So, I'm wondering whether the above helm command work well or not if we redirect this URL to another URL.

kota2and3kan (commented):
Thank you for the update!
I left one question. Please take a look when you have time!

_config.yml
josh-wong: My mistake! I just checked the [`.github/workflows/helm_charts_scalar.yml`](https://github.com/scalar-labs/helm-charts/blob/main/.github/workflows/helm_charts_scalar.yml) file and noticed that the base URL points to the content for the `charts` folder rather than the `docs` folder.

Instead, if we point https://scalar-labs.github.io/helm-charts/docs/ to the [product selection page on the new Scalar docs site](https://developers.scalar-labs.com/docs/scalar-helm-charts-product-selection), do you see any issues with that?

kota2and3kan: Thank you for the confirmation!


> Instead, if we point https://scalar-labs.github.io/helm-charts/docs/ to the product selection page on the new Scalar docs site, do you see any issues with that?

I think it's OK because the `helm` command does not use the `docs/` path.

---

For your reference, the `helm` command accesses the following three URLs.

1. Root `/` of the helm repository when it adds a new repository or resolves some dependencies.
   ```
   https://scalar-labs.github.io/helm-charts
   ```
   * For example, we specify this URL to the [dependency configuration](https://github.com/scalar-labs/helm-charts/blob/scalardb-2.6.1/charts/scalardb/Chart.yaml#L20) in the `Chart.yaml`.
1. The `/index.yaml` when it searches some metadata of each chart.
   ```
   https://scalar-labs.github.io/helm-charts/index.yaml
   ```
1. GitHub releases' URL that provides actual chart files (`*.tgz` files).
   ```
   https://github.com/scalar-labs/helm-charts/releases/download/<RELEASE_TAG>/<CHART_NAME>-<CHART_VERSION>.tgz
   ```
   * For example, you can see the `*.tgz` file path in the [index.yaml](https://github.com/scalar-labs/helm-charts/blob/main/index.yaml#L363-L364) file.

So, we should not redirect these URLs.

josh-wong: Understood. Thank you for the detailed explanation! 

I will rework the redirect to ensure only pages within https://scalar-labs.github.io/helm-charts/docs/ are redirected.

josh-wong: @kota2and3kan I've found a way to implement a redirect for the GitHub Pages deployment in this repository. This redirect does the following:

- Does not redirect users who visit https://scalar-labs.github.io/helm-charts/.
- Redirects users who visit the docs at https://scalar-labs.github.io/helm-charts/docs/.

To test this redirect method, I did the following:

- Forked this repository.
- Moved the three files in this PR from the `docs` folder to the root folder.
- Confirmed that the deployed site from my forked repository _did not_ redirect me when visiting the **Scalar Helm Charts pages**.
- Confirmed that the deployed site from my forked repository _did_ redirect me when visiting the **docs for Scalar Helm Charts** to the [Scalar Helm Charts product selection](https://developers.scalar-labs.com/docs/scalar-helm-charts-product-selection) page.

PTAL, and let me know if you have any questions!

kota2and3kan: @josh-wong  
Thank you for the updates and testing!
The behaviors that you mentioned and tested look good to me!

kota2and3kan (approved):
LGTM! Thank you!

PR #235, opened 2023-08-08
[scalardb-cluster] Fix environment variable name for log level

kota2and3kan:
This PR fixes the environment variable name that we use for setting of log level.
In the `log4j2.properties` file, ScalarDB Cluster uses the environment variable `SCALAR_DB_CLUSTER_LOG_LEVEL`.
https://github.com/scalar-labs/scalardb-cluster/blob/v3.10.1/node/conf/log4j2.properties

So, we must use this environment variable name in the Helm Chart for ScalarDB Clsuter.

Please take a look!


brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #236, opened 2023-08-17
Change link for "Getting Started with ScalarDB"

josh-wong:
This PR changes the link for the "Getting Started with ScalarDB" doc, which we updated in https://github.com/scalar-labs/scalardb/pull/990. The previous doc no longer exists, so that's why we need to update this link.


feeblefakie (approved):
LGTM!

PR #237, opened 2023-09-14
Update supported Kubernetes version

kota2and3kan:
This PR updates the supported Kubernetes version.

At this time, each managed Kubernetes service supports the following versions.

* [EKS](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html)
  * v1.23 - v1.27
* [AKS](https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions)
  * v1.25 - v1.28

So, I updated the supported Kubernetes versions that the Scalar Helm Chart supports based on the above versions.

Also, we test these new versions in our CI as follows.
https://github.com/scalar-labs/helm-charts/actions/runs/6155343603

Please take a look!

### DIFF
--- a/docs/how-to-deploy-scalar-products.md
+++ b/docs/how-to-deploy-scalar-products.md
@@ -25,7 +25,7 @@ You must prepare a Kubernetes cluster for the deployment of Scalar products. If 
 
 * [scalar-labs/scalar-kubernetes](https://github.com/scalar-labs/scalar-kubernetes/blob/master/README.md)
 
-You must prepare a Kubernetes with supported versions. See [Supported Kubernetes versions](https://github.com/scalar-labs/helm-charts#supported-kubernetes-versions) for the versions that Scalar Helm Chart supports.
+You must prepare a supported version of Kubernetes. For versions that Scalar Helm Charts supports, see [Supported Kubernetes versions](https://github.com/scalar-labs/helm-charts#supported-kubernetes-versions).
 
 ### Prepare a database (ScalarDB, ScalarDL Ledger, ScalarDL Auditor)
 
--- a/docs/how-to-deploy-scalar-products.md
+++ b/docs/how-to-deploy-scalar-products.md
@@ -25,6 +25,8 @@ You must prepare a Kubernetes cluster for the deployment of Scalar products. If 
 
 * [scalar-labs/scalar-kubernetes](https://github.com/scalar-labs/scalar-kubernetes/blob/master/README.md)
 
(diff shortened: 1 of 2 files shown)


kota2and3kan:
@josh-wong 
Thank you for the review and question!

> Since visitors to the ScalarDB and ScalarDL docs sites won't be able to see which versions are supported (the root README isn't added to the docs sites), should we create a doc that shows these supported Kubernetes versions?

Ah, I see. That's right. Users are not able to see this README in doc site.

I think we should add this `supported Kubernetes versions` information in any documents on the doc site side.

> Or is that information noted elsewhere that is easy for users to reference (for example, in the managed Kubernetes services when users set up Scalar Helm Charts)?

I think there is no document that describes the `supported Kubernetes versions` in any other documents.

---

I checked some documents on the doc site. As a result, I think it might be better to add a description (link) about `Supported Kubernetes Versions` in the following document.
https://github.com/scalar-labs/helm-charts/blob/main/docs/how-to-deploy-scalar-products.md

This is because users can see this document from both ScalarDB doc site and ScalarDL doc site, and this document already explains `Prerequisites`. (I think the `supported version` is one of the prerequisites.)

I just added the link in e88e205ed93bbcf4df2700d85cc965c8448c0e9f.

What do you think?

docs/how-to-deploy-scalar-products.md
```diff
@@ -25,6 +25,8 @@ You must prepare a Kubernetes cluster for the deployment of Scalar products. If
 
 * [scalar-labs/scalar-kubernetes](https://github.com/scalar-labs/scalar-kubernetes/blob/master/README.md)
 
+You must prepare a Kubernetes with supported versions. See [Supported Kubernetes versions](https://github.com/scalar-labs/helm-charts#supported-kubernetes-versions) for the versions that Scalar Helm Chart supports.
```
josh-wong, line 28:
```suggestion
You must prepare a supported version of Kubernetes. For versions that Scalar Helm Charts supports, see [Supported Kubernetes versions](https://github.com/scalar-labs/helm-charts#supported-kubernetes-versions).
```

kota2and3kan: Looks good! Thank you for your suggestion!

kota2and3kan:
@josh-wong 
Thank you for your suggestion! I applied it. So, please take a look again when you have time!

> However, I understand that we should limit needing to make the same change in multiple places.

Yes. I want to avoid duplicated documents to reduce maintenance costs.
I think we should consider re-organizing Scalar Helm Chart docs in the future.

feeblefakie (approved):
LGTM! Thank you!

PR #238, opened 2023-09-14
Fix broken link in configure ScalarDB Cluster doc

kota2and3kan:
This PR fixes the broken link in the document.
The linked document was updated before.
Please take a look!

---

@josh-wong 
I have one question about the document.

We can see the same document in both GitHub and Scalar Doc Site as follows.

- https://github.com/scalar-labs/scalardb-cluster/blob/main/docs/scalardb-cluster-configurations.md
- https://scalardb.scalar-labs.com/docs/latest/scalardb-cluster/scalardb-cluster-configurations/

In this case, which URL should we set in the document if we want to link other documents from a document?

In this PR I used GitHub's URL. Is it OK to use GitHub's URL when I link some other documents from one document?

If we should use the Scalar Doc Site URL, I will update this PR.

### DIFF
--- a/docs/configure-custom-values-scalardb-cluster.md
+++ b/docs/configure-custom-values-scalardb-cluster.md
@@ -17,7 +17,7 @@ scalardbCluster:
 
 ### Database configurations
 
-You must set `scalardbCluster.scalardbClusterNodeProperties`. Please set `scalardb-cluster-node.properties` to this parameter. For more details on the configurations of ScalarDB Cluster, see [Configurations](https://github.com/scalar-labs/scalardb-cluster/blob/main/docs/index.md#configurations).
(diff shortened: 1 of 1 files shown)


josh-wong (approved):
For now, please use the GitHub URL (as you've done in this PR). 

It's a bit tricky because, currently, if we use a URL for the docs site, then users will be taken to the `latest` version of docs if they are looking at a different version of docs, like `3.8`.

I'm still considering how to avoid this type of behavior by keeping users on the docs site and within the version of docs they are reading🙇🏻‍♂️

kota2and3kan:
@josh-wong 
Thank you for your review and answer!

> For now, please use the GitHub URL (as you've done in this PR). 
> 
> It's a bit tricky because, currently, if we use a URL for the docs site, then users will be taken to the `latest` version of docs if they are looking at a different version of docs, like `3.8`.

An, I see. It seems that it's a bit complex...
Anyway, I understand. I will use GitHub's URL. Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #242, opened 2023-11-07
Add new chart for ScalarDB Analytics with PostgreSQL

kota2and3kan:
## Description

This PR adds a new helm chart for `ScalarDB Analytics with PostgreSQL`!

Deploying `ScalarDB Analytics with PostgreSQL` on the Kubernetes environment manually takes a bit of time and effort. Instead of that, this chart can help users to deploy ScalarDB Analytics with PostgreSQL on the Kubernetes environment.

## Related issues and/or PRs

N/A

## Changes made

* Add a new chart (new resource manifests).
* Add new test for this new chart in the CI.
* Add new documents that describes how to use this new chart.
* Update existing documents to add the descriptions of this new chart.

## Checklist

### Regarding getting started guide

I will create `Getting started guide` in another PR after related issue https://github.com/scalar-labs/scalardb-analytics-postgresql/issues/44 fixed on `ScalarDB Analytics with PostgreSQL` side.

### Overview of what you need to run `ScalarDB Analytics with PostgreSQL`

In arbitrary platforms including other environments than Kubernetes, when you run `ScalarDB Analytics with PostgreSQL`, you need to run the following two steps.

0. Before running `ScalarDB Analytics with PostgreSQL` (there are several backend databases).

   ```console
   +-------------------+
   | Backend databases |
   +-------------------+
   ```

1. Run `ScalarDB Analytics with PostgreSQL` container as a first step.

   ```console
   +-------------------+      +------------------------------------+
   | Backend databases | <--- | ScalarDB Analytics with PostgreSQL |
   +-------------------+      +------------------------------------+
   ```

2. Run Schema Importer against the ScalarDB Analytics with PostgreSQL container to load some objects into PostgreSQL as a second step.

   ```console
   +-------------------+      +------------------------------------+                       +-----------------+
   | Backend databases | <--- | ScalarDB Analytics with PostgreSQL | <---(load objects)--- | Schema Importer |
   +-------------------+      +------------------------------------+                       +-----------------+
   ```

These are the general way to run `ScalarDB Analytics with PostgreSQL`.

### How this chart deploy `ScalarDB Analytics with PostgreSQL`

In this chart, it combines `ScalarDB Analytics with PostgreSQL` and `Schema Importer` into one Pod. And, run `Schema Importer` automatically in the pod.

0. Before running `ScalarDB Analytics with PostgreSQL` pod (there are several backend databases).

   ```console
   +-------------------+
   | Backend databases |
   +-------------------+
   ```

1. Run `ScalarDB Analytics with PostgreSQL` pod.

   ```console
                              +--[Pod]-----------------------------------------+
                              |                                                |
                              |  +------------------------------------+        |
                              |  | ScalarDB Analytics with PostgreSQL |        |
   +-------------------+      |  +------------------------------------+        |
   | Backend databases | <--- |                                                |
   +-------------------+      |  +-----------------+                           |
                              |  | Schema Importer |                           |
                              |  +-----------------+                           |
                              |                                                |
                              +------------------------------------------------+
   ```

2. Automatically, run Schema Importer against the `ScalarDB Analytics with PostgreSQL` container in the pod.

   ```console
                              +--[Pod]-----------------------------------------+
                              |                                                |
                              |  +------------------------------------+        |
                              |  | ScalarDB Analytics with PostgreSQL | <---+  |
   +-------------------+      |  +------------------------------------+     |  |
   | Backend databases | <--- |                                             |  |
   +-------------------+      |  +-----------------+                        |  |
                              |  | Schema Importer | ---(load objects)------+  |
                              |  +-----------------+                           |
                              |                                                |
                              +------------------------------------------------+
   ```

3. If Schema Importer fails because PostgreSQL is not started yet, `entrypoint.sh` retries to run the `Schema Importer` several times (10 times by default).

   ```console
                              +--[Pod]-----------------------------------------+
                              |                                                |
                              |  +------------------------------------+        |
                              |  | ScalarDB Analytics with PostgreSQL | <---+  |
   +-------------------+      |  +------------------------------------+     |  |
   | Backend databases | <--- |                                             |  |
   +-------------------+      |  +-----------------+                        |  |
                              |  | Schema Importer | ---(retry if failed)---+  |
                              |  +-----------------+                           |
                              |                                                |
                              +------------------------------------------------+
   ```

4. After Schema Importer succeeds, the Schema Importer container will sleep endlessly (run the `sleep inf` command).

   ```console
                              +--[Pod]-----------------------------------------+
                              |                                                |
                              |  +------------------------------------+        |
                              |  | ScalarDB Analytics with PostgreSQL |        |
   +-------------------+      |  +------------------------------------+        |
   | Backend databases | <--- |                                                |
   +-------------------+      |  +-----------------------------+               |
                              |  | Schema Importer (sleep inf) |               |
                              |  +-----------------------------+               |
                              |                                                |
                              +------------------------------------------------+
   ```

## Release notes

Add a new chart for ScalarDB Analytics with PostgreSQL. By using this new chart, you can deploy ScalarDB Analytics with PostgreSQL on the Kubernetes environment.


charts/scalardb-analytics-postgresql/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
+name: scalardb-analytics-postgresql
+description: ScalarDB Analytics with PostgreSQL
+type: application
+version: 1.0.0-SNAPSHOT
+appVersion: 3.10.2
```
kota2and3kan, line 6: At this time, ScalarDB Analytics with PostgreSQL does not release the `SNAPSHOT` version. So, I set the latest stable version.

However, we are working on releasing the `SNAPSHOT` version on the ScalarDB Analytics with PostgreSQL side. In the future, we will set the `SNAPSHOT` version here.

charts/scalardb-analytics-postgresql/README.md
```diff
@@ -0,0 +1,54 @@
+# scalardb-analytics-postgresql
```
kota2and3kan, line 1: This file is automatically generated based on the `values.yaml` file.

charts/scalardb-analytics-postgresql/ci/scalardb-analytics-postgresql-ct-values.yaml
```diff
@@ -0,0 +1,10 @@
+scalardbAnalyticsPostgreSQL:
+  databaseProperties: |
+    scalar.db.storage=jdbc
+    scalar.db.contact_points=jdbc:postgresql://postgresql.default.svc.cluster.local:5432/postgres
+    scalar.db.username=postgres
+    scalar.db.password=postgres
+
+schemaImporter:
+  namespaces:
+    - ct
```
kota2and3kan, line 10: This file is a custom values file for the testing in the CI.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/configmap.yaml
```diff
@@ -0,0 +1,48 @@
...
+    {{- toYaml .Values.scalardbAnalyticsPostgreSQL.databaseProperties | nindent 4 }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "scalardb-analytics-postgresql.fullname" . }}-entrypoint-shell
+  namespace: {{ .Release.Namespace }}
+data:
+  entrypoint.sh: |
+    #!/bin/bash
```
kota2and3kan, line 18: At this time, I create the `entorypoint.sh` on the helm chart side. Schema Importer container mounts this file and runs it as an entrypoint. 

The main purpose of this shell is to implement the `retry` process for Schema Importer.

We need this file on the helm chart side to run the existing stable versions (v3.10) images.

However, I will create this `entrypoint.sh` on the Schema Imorter container image side in the future. After that (maybe after v3.11), we can use the `etntrypoint.sh` that is included in the container image.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
```diff
@@ -0,0 +1,154 @@
...
+          image: "{{ .Values.scalardbAnalyticsPostgreSQL.image.repository }}:{{ .Values.scalardbAnalyticsPostgreSQL.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.scalardbAnalyticsPostgreSQL.image.pullPolicy }}
+          ports:
+            - containerPort: 5432
+          env:
+          - name: POSTGRES_DB
+            value: "{{ .Values.scalardbAnalyticsPostgreSQL.postgresql.databaseName }}"
+          envFrom:
+          - secretRef:
+              name:  "{{ .Values.scalardbAnalyticsPostgreSQL.postgresql.secretName }}"
```
kota2and3kan, line 53: We set the superuser's password of PostgreSQL via a Secret resource.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
```diff
@@ -0,0 +1,154 @@
...
+          livenessProbe:
+            exec:
+              command:
+                - psql
+                - -d
+                - "{{ .Values.scalardbAnalyticsPostgreSQL.postgresql.databaseName }}"
+                - -U
+                - postgres
+                - -c
+                - "SELECT 1"
```
kota2and3kan, line 85: I use `SELECT 1` for the health checking.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
```diff
@@ -0,0 +1,154 @@
...
+            failureThreshold: 3
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+        - name: schema-importer
```
kota2and3kan, line 90: As I mentioned in the PR description, this chart run `Schema Importer` as a sidecar.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
```diff
@@ -0,0 +1,154 @@
...
+              name: entrypoint-shell-volume
+              subPath: entrypoint.sh
+          {{- with .Values.scalardbAnalyticsPostgreSQL.extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+      volumes:
+        - configMap:
+            defaultMode: 0440
+            name: {{ include "scalardb-analytics-postgresql.fullname" . }}-database-properties
+          name: database-properties-volume
```
kota2and3kan, line 135: The `ScalarDB Analytics with PostgreSQL` container and `Schema Importer` container share the same `database.properties` file.

charts/scalardb-analytics-postgresql/values.schema.json
```diff
@@ -0,0 +1,239 @@
+{
```
kota2and3kan, line 1: This file is automatically generated based on the `values.yaml` file.

charts/scalardb-analytics-postgresql/values.yaml
```diff
@@ -0,0 +1,163 @@
...
+        port: 5432
+        # -- PostgreSQL k8s internal port
+        targetPort: 5432
+        # -- PostgreSQL protocol
+        protocol: TCP
+
+  # -- PodSecurityContext holds pod-level security attributes and common container settings.
+  podSecurityContext:
+    # -- To work ScalarDB Analytics with PostgreSQL properly, you must set "201" to "podSecurityContext.fsGroup".
+    fsGroup: 201
```
kota2and3kan, line 76: To run the `entorypoint.sh` by `scalar (UID=201)` user in the Schema Importer container, we have to mount the `entrypoint.sh` file with `201:201` configuration as a file owner configuration. So, we have to set `fsGroup=201` here.

charts/scalardb-analytics-postgresql/values.yaml
```diff
@@ -0,0 +1,163 @@
...
+  # -- Setting security context at the pod applies those settings to all containers in the pod.
+  securityContext:
+    # -- Capabilities (specifically, Linux capabilities), are used for permission management in Linux. Some capabilities are enabled by default.
+    capabilities:
+      drop:
+        - ALL
+    # -- Containers should be run as a non-root user with the minimum required permissions (principle of least privilege).
+    runAsNonRoot: true
+    # -- The PostgreSQL official image use the "postgres (UID=999)" user by default.
+    runAsUser: 999
```
kota2and3kan, line 89: The PostgreSQL container image sets non-root user `postgres` with `UID=999`. To run the container with the non-root user properly, we have to set `UID=999` here.
https://github.com/scalar-labs/docker/blob/main/jdk-postgres/8-15/Dockerfile#L14

charts/scalardb-analytics-postgresql/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
+name: scalardb-analytics-postgresql
+description: ScalarDB Analytics with PostgreSQL
+type: application
+version: 1.0.0-SNAPSHOT
+appVersion: 3.10.3
```
kota2and3kan, line 6: At this time, ScalarDB Analytics with PostgreSQL does not release the `SNAPSHOT` version. So, I set the latest stable version.

However, we are working on releasing the `SNAPSHOT` version on the ScalarDB Analytics with PostgreSQL side. In the future, we will set the `SNAPSHOT` version here.

docs/configure-custom-values-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,179 @@
+# Configure a custom values file for ScalarDB Analytics with PostgreSQL
+
+This document explains how to create your custom values file for the ScalarDB Analytics with PostgreSQL chart. For details on the parameters, see the [README](https://github.com/scalar-labs/helm-charts/blob/main/charts/scalardb-analytics-postgresql/README.md) of the ScalarDB Analytics with PostgreSQL chart.
+
+## Required configurations
+
+### Database configurations
+
+You must set `scalardbAnalyticsPostgreSQL.databaseProperties`. Please set `database.properties` to this parameter. For more details on the configurations of ScalarDB, see [ScalarDB Configurations](https://github.com/scalar-labs/scalardb/blob/master/docs/configurations.md).
```
komamitsu, line 9: Sorry, I'm a bit confused with this. Does it mean to set `scalardbAnalyticsPostgreSQL.databaseProperties` to the content of `database.properties`?

kota2and3kan: @komamitsu  
Thank you for the question!

It means that the parameter `scalardbAnalyticsPostgreSQL.databaseProperties` should include the properties with the same syntax as the `database.properties` of ScalarDB.

For example, if you want to read two tables in two backend databases via ScalarDB Analytics with PostgreSQL, you need to set the configuration for `Multi-storage transactions` in `scalardbAnalyticsPostgreSQL.databaseProperties`.

* Read two tables data via ScalarDB Analytics with PostgreSQL.

  ```console
                                                  +-----------------------------+
                                            +---> | MySQL ("customer" schema)   |
                                            |     +-----------------------------+
  +------------------------------------+    |
  | ScalarDB Analytics with PostgreSQL | ---+
  +------------------------------------+    |
                                            |     +-----------------------------+
                                            +---> | PostgreSQL ("order" schema) |
                                                  +-----------------------------+
  ```

* Custom values file (setting of `scalardbAnalyticsPostgreSQL.databaseProperties`) in the above case.

  ```yaml
  scalardbAnalyticsPostgreSQL:
    databaseProperties: |
      scalar.db.storage=multi-storage
      scalar.db.multi_storage.storages=storage0,storage1
      
      # Storage 0
      scalar.db.multi_storage.storages.storage0.storage=jdbc
      scalar.db.multi_storage.storages.storage0.contact_points=jdbc:mysql://mysql-scalardb.default.svc.cluster.local:3306/
      scalar.db.multi_storage.storages.storage0.username=root
      scalar.db.multi_storage.storages.storage0.password=mysql
      
      # Storage 1
      scalar.db.multi_storage.storages.storage1.storage=jdbc
      scalar.db.multi_storage.storages.storage1.contact_points=jdbc:postgresql://postgresql-scalardb.default.svc.cluster.local:5432/postgres
      scalar.db.multi_storage.storages.storage1.username=postgres
      scalar.db.multi_storage.storages.storage1.password=postgres
      
      scalar.db.multi_storage.namespace_mapping=customer:storage0,order:storage1
      scalar.db.multi_storage.default_storage=storage1

  schemaImporter:
    namespaces:
      - customer
      - order
  ```

komamitsu: Thanks for the details. I think we are on the same page. So, I think there might be a room to make the sentence a bit clearer, but I'll leave it to @josh-wong 😅

docs/configure-custom-values-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+### PostgreSQL superuser password configuration (optional based on your environment)
+
+You can specify the secret name that includes superuser password for PostgreSQL.
+
+```yaml
+scalardbAnalyticsPostgreSQL:
+  postgresql:
+  secretName: scalardb-analytics-postgresql-superuser-password
```
komamitsu, line 157: I guess this corresponds to `scalardbAnalyticsPostgreSQL.postgresql.secretName`. If so, this line should be indented?
```suggestion
    secretName: scalardb-analytics-postgresql-superuser-password
```

kota2and3kan: Ah, yes. You are right!
I will apply it. Thank you for pointing it out!

docs/configure-custom-values-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+### PostgreSQL database name configuration (optional based on your environment)
+
+You can specify the database name that you create in PostgreSQL. Schema Importer creates some object such a view of ScalarDB Analytics with PostgreSQL in this database.
+
+```yaml
+scalardbAnalyticsPostgreSQL:
+  postgresql:
+  databaseName: scalardb
```
komamitsu, line 147:
```suggestion
    databaseName: scalardb
```
Same as below?

kota2and3kan: Yeah, right!
Sorry, I overlooked it...
Thank you for pointing it out!

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
```diff
@@ -0,0 +1,155 @@
...
+          - secretRef:
+              name:  "{{ .Values.scalardbAnalyticsPostgreSQL.postgresql.secretName }}"
+          {{- if .Values.scalardbAnalyticsPostgreSQL.secretName }}
+          - secretRef:
+              name: "{{ .Values.scalardbAnalyticsPostgreSQL.secretName }}"
+          {{- end }}
+          volumeMounts:
+            - mountPath: /etc/database.properties
+              name: database-properties-volume
+              subPath: database.properties
```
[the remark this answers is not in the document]
kota2and3kan: Thank you for the suggestion!
We don't update this configuration file in the pod.
So, we can make this file read-only.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
```diff
@@ -0,0 +1,155 @@
...
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/scalardb-analytics-postgresql/configmap.yaml") . | sha256sum }}
+        {{- if .Values.scalardbAnalyticsPostgreSQL.podAnnotations }}
+        {{- toYaml .Values.scalardbAnalyticsPostgreSQL.podAnnotations | nindent 8 }}
+        {{- end }}
+      labels:
+        {{- include "scalardb-analytics-postgresql.selectorLabels" . | nindent 8 }}
+    spec:
+      restartPolicy: Always
```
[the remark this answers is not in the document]
kota2and3kan: Ah, I see. Thank you for the suggestion!
There is no special meaning that I specify here explicitly.
So, I will remove it.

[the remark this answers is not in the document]
kota2and3kan: Fixed in 5434a0b11c26d6a1429cdeebc5262ccd5dc21f0d.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
```diff
@@ -0,0 +1,155 @@
...
+            - --host
+            - localhost
+            - --port
+            - "5432"
+            - --database
+            - "{{ .Values.scalardbAnalyticsPostgreSQL.postgresql.databaseName }}"
+            - --user
+            - postgres
+            - --password
+            - "{{ .Values.scalardbAnalyticsPostgreSQL.postgresql.password }}"
```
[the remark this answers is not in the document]
kota2and3kan: @superbrothers (cc: @choplin )

Thank you for your suggestion! It's a very important point!
In conclusion, we can remove this `--password` option. So, I just removed it in ec0c4b5a554b99a15c0b0a6e894becb04b1ec97c.

After I checked this point, I found that we don't need to set the `--password` option if we run `Schema Importer` against `localhost`. This is because the default `pg_hba.conf` allows connections from `localhost` (i.e., connections from the sidecar container via `localhost`) without a password as follows.

```console
host    all             all             127.0.0.1/32            trust
```

Note: Of course, the connections from outside of the pod still need password authentication.

So, we can remove this `--password` option if we run `Schema Importer` in the sidecar container.

---

For your information, we will update `Schema Importer` to get a password from an environment variable in the future version.
https://github.com/scalar-labs/scalardb-analytics-postgresql/issues/47

Also, there is a possibility that the default value of pg_hba.conf will be changed. However, even in such cases, we can configure pg_hba.conf file ourselves because the PostgreSQL container image provided such feature `-c 'config_file=/etc/postgresql/postgresql.conf'`.

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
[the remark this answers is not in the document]
kota2and3kan: I also made other mounted files read-only in 9f76818e65d7ddd731c6c7a5c1a237565f82ab05 because we don't update those files in the pod.

kota2and3kan:
@superbrothers 
Thank you for your review!
I applied some suggestions and updated some manifests based on the feedback.
Please take a look again when you have time!

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
superbrothers: I prefer to use `env` because I don't know the key name that should be included in the secret from envFrom (although it is case by case).

superbrothers (approved):

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/configmap.yaml
brfrn169: So, this runs Schema Importer every time the pod starts?

docs/configure-custom-values-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,179 @@
...
+```yaml
+scalardbAnalyticsPostgreSQL:
+  databaseProperties: |
+    scalar.db.contact_points=localhost
+    scalar.db.username=${env:SCALAR_DB_USERNAME:-}
+    scalar.db.password=${env:SCALAR_DB_PASSWORD:-}
+    scalar.db.storage=cassandra
+```
+
+### Namespaces configurations
```
brfrn169, line 20: This might cause confusion between ScalarDB's namespace and Kubernetes's namespace.

brfrn169 (commented):
Overall, LGTM. Left several minor comments. Please take a look when you have time!

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
kota2and3kan: Ah, I see. In this case, the environment variable name is fixed `POSTGRES_PASSWORD`. It depends on the [PostgreSQL official container image](https://hub.docker.com/_/postgres).

In other words, users cannot use arbitrary environment variable names, and there is no special reason that I use `envFrom` here.

So, I will update to use `env`. Thank you for your suggestion!

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/configmap.yaml
kota2and3kan: Yes. That's right. For example, we run Schema Importer in the following cases:

* Deploy pods.
* Pods restart for some reason.
* Scale out the pods.

docs/configure-custom-values-scalardb-analytics-postgresql.md
kota2and3kan: Ah, I see. I agree with that concern.
I will update the documents.
Thank you for pointing it out!

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/deployment.yaml
kota2and3kan: Fixed in 8af1a27313f53b9c07d1e36a76d6127f210c38bc.

docs/configure-custom-values-scalardb-analytics-postgresql.md
kota2and3kan: I updated the document to describe the `namespace` is the `database namespace` in this context explicitly in 831fa610740cbced854e3409c22c4c8afd4a25d1.

docs/configure-custom-values-scalardb-analytics-postgresql.md
josh-wong: @kota2and3kan I'm still a bit confused since I'm reading it two different ways. I've provided two suggestions below, so please help me understand which one is accurate🙇🏻‍♂️

> [!NOTE]
> 
> Also, I changed `database.properties` to `ScalarDB .properties file` because I think the file name can be whatever the user chooses (for example, `database.properties`, `scalardb.properties`, etc.). Is that OK?

```suggestion
You must set the `scalardbAnalyticsPostgreSQL.databaseProperties` parameter by following the same syntax that you use to configure the ScalarDB .properties file. For details about configurations, see [ScalarDB Configurations](https://github.com/scalar-labs/scalardb/blob/master/docs/configurations.md).
```

Or, if the contents of `scalardbAnalyticsPostgreSQL.databaseProperties` and the ScalarDB .properties file need to match exactly, the following might be better:

```suggestion
You must set the `scalardbAnalyticsPostgreSQL.databaseProperties` parameter to the same configurations that you have set in the ScalarDB .properties file. For details about configurations, see [ScalarDB Configurations](https://github.com/scalar-labs/scalardb/blob/master/docs/configurations.md).
```

kota2and3kan: @josh-wong  

---

> I've provided two suggestions below, so please help me understand which one is accurate🙇🏻‍♂️

We don't need to exactly match the `database.properties` of ScalarDB (basically, ScalarDB Cluster) and ScalarDB Analytics with PostgreSQL. This is because ScalarDB Analytics with PostgreSQL configuration (`database.properties`) requires only database information that users want to read data via ScalarDB Analytics.

For example, users run the OLTP workload via ScalarDB against 5 databases. However, if they want to run the OLAP workload via ScalarDB Analytics with PostgreSQL against only 3 databases, it is enough that the `database.properties` for ScalarDB Analytics includes configurations for only 3 databases.

```
                                         +------+
                                   +---> | DB 1 |
                                   |     +------+
                                   |
                                   |     +------+
                                   +---> | DB 2 | <---+
                                   |     +------+     |
                                   |                  |
+-------------+    +----------+    |     +------+     |    +------------------------------------+      +-------------+
| OLTP client |--->| ScalarDB | ---+---> | DB 3 | <---+--- | ScalarDB Analytics with PostgreSQL | <--- | OLAP client |
+-------------+    +----------+    |     +------+     |    +------------------------------------+      +-------------+
                                   |                  |
                                   |     +------+     |
                                   +---> | DB 4 | <---+
                                   |     +------+
                                   |
                                   |     +------+
                                   +---> | DB 5 |
                                         +------+
```

In this case, the `database.properties` of ScalarDB and ScalarDB Analytics is different. So, the first one looks good to me.

Also, I think it would be better to add a bit of detail to represent the condition `exact matching is not necessary`.

```suggestion
To access databases via ScalarDB Analytics with PostgreSQL, you must set the `scalardbAnalyticsPostgreSQL.databaseProperties` parameter by following the same syntax that you use to configure the ScalarDB .properties file. For details about configurations, see [ScalarDB Configurations](https://github.com/scalar-labs/scalardb/blob/master/docs/configurations.md).
```

What do you think?

---

> Also, I changed `database.properties` to `ScalarDB .properties file` because I think the file name can be whatever the user chooses (for example, `database.properties`, `scalardb.properties`, etc.). Is that OK?

From the technical perspective, you are right, and I think it's OK to use the term `ScalarDB .properties file`. However, I have one concern. In other ScalarDB documents, I think they state `database.properties` or `scalardb.properties`. In other words, it is not consistent with other documents.

So, there is a possibility that users are confused about `Are the ScalarDB .properties file and database.properties different?`. For example, if users first see ScalarDB documents and then see this document, it causes such confusion I think.

Of course, I think mixing  `database.properties` and `scalardb.properties` in several documents is not good... So, we should unify the terming to `ScalarDB .properties file` in the whole documents of ScalarDB.

As a first step of that, I agree with using the term `ScalarDB .properties file` in this document. But, it affects all ScalarDB documents. So, I think we should discuss terming of properties file with other product members first.

Also, it might be better to update this term in another PR (in another task/story) such as `Unify properties file names in all documents`.

What do you think?

kota2and3kan:
@brfrn169  
Thank you for your review!
I updated the documents and left an answer to your question.
Please take a look again when you have time!

docs/configure-custom-values-scalardb-analytics-postgresql.md
josh-wong: > In this case, the database.properties of ScalarDB and ScalarDB Analytics is different. So, the first one looks good to me.
> 
> Also, I think it would be better to add a bit of detail to represent the condition exact matching is not necessary.

@kota2and3kan Thank you for confirming!🙇🏻‍♂️  Let's use the following, with the phrase for the properties file (`database.properties`) for consistency across other docs in this PR.

```suggestion
To access databases via ScalarDB Analytics with PostgreSQL, you must set the `scalardbAnalyticsPostgreSQL.databaseProperties` parameter by following the same syntax that you use to configure the `database.properties` file. For details about configurations, see [ScalarDB Configurations](https://github.com/scalar-labs/scalardb/blob/master/docs/configurations.md).
```

> Also, it might be better to update this term in another PR (in another task/story) such as Unify properties file names in all documents.

I agree. I've only recently found out that the user can specify the name of the `.properties`, so I'd like to note that the file name may vary depending on the user's environment in the future.

As you mentioned, let's determine the exact phrase to use in another PR and across docs. Maybe we can do that after we consolidate docs into one repository to ensure consistency💪

kota2and3kan: @josh-wong  
> Let's use the following, with the phrase for the properties file (`database.properties`) for consistency across other docs in this PR.

Looks good. I will apply it.
Thank you for updating your suggestion!

> As you mentioned, let's determine the exact phrase to use in another PR and across docs. Maybe we can do that after we consolidate docs into one repository to ensure consistency💪

Sound good. Let's discuss and update documents after we unify the docs into one repository!

feeblefakie (approved):
LGTM! Thank you!

charts/scalardb-analytics-postgresql/templates/scalardb-analytics-postgresql/configmap.yaml
brfrn169: Thank you for the response. I have one more question: why is it necessary to include the Schema Importer in the pod? I'm considering the possibility of running the Schema Importer separately, outside of the pod.

kota2and3kan: @brfrn169  
Thank you for your question!

In conclusion, I want to deploy ScalarDB Analytics with PostgreSQL as a `Stateless` workload to reduce maintenance costs and complex configurations, at the moment. This is why I run the Schema Importer as a sidecar in the pod.

For example, I want to avoid manually running the Schema Importer in the following cases.

* Existing pods crash or restart
* Scale pods (new pods start)

## Challenges

ScalarDB Schema Loader and ScalarDL Schema Loader create database schemas (i.e., create some tables or objects) on the backend database side. In other words, those objects are persisted by the backend database. So, basically, ScalarDB/ScalarDL has no state in themselves.

However, Schema Importer creates some objects (e.g., foreign servers, extensions, and views) on the ScalarDB Analytics with PostgreSQL side. In other words, strictly, ScalarDB Analytics with PostgreSQL has states. It's a `Stateful` workload.

So, if the pod is restarted for some reason, the loaded objects are lost. In this case, we have to re-run Schema Importer to re-load all objects on the PostgreSQL.

As well as the pods restart, we have to run the Schema Importer if we scale out (add a new pod) to create some objects in the new pod (in the new PostgreSQL). This is because each pod (each PostgreSQL) has objects (e.g., views) respectively.

## Solution 1 (make it `Stateful` workload)

To address the above challenges, we can deploy ScalarDB Analytics with PostgreSQL as a `Stateful` workload by using `StatefulSet` which is one of the Kubernetes resources.

In this case, the objects (foreign server, extension, and views) are stored and persisted in the PV (persistent volume) which is attached to the pod.

So, we don't need to re-run Schema Importer if the pods crash/restart. However, we still have to run Schema Importer manually when we scale out pods.

Also, in this solution, we have to use a bit more complex configurations for instance `StatefulSet` and `PersistentVolume`, rather than we deploy it as a `Stateless` workload by using `Deployment`. It takes maintenance costs.

In addition, in this case, we have to consider the backup/restore of the `PersistentVolume`. It increases operation costs.

So, I want to avoid this solution if I can.

## Solution 2 (run Schema Importer manually every time)

This is a simple (but not easy) solution. We can deploy ScalarDB Analytics with PostgreSQL as a `Stateless` workload with `Deployment`, and run Schema Importer manually every time we need.

However, this solution increases the operation costs on the user side. And, this solution cannot take advantage of the self-healing (automatically pod restart when some failure occurs) feature of Kubernetes well.

## Solution 3 (run Schema Importer as a sidecar / our choice)

To resolve the challenges with the smallest additional costs, we decided to run Schema Importer as a sidecar in the pod startup step.

In this case, the pod runs Schema Importer automatically when pods crash/restart or a new pod is added. So, we can avoid manually running Schema Importer.

Also, in this case, we don't need to store/persist some objects in the `PersistentVolume` because the pod runs Schema Importer every time on its startup. We can avoid additional costs to maintain the `Stateful` workload.

This is why I decided to run Schema Importer as a sidecar.

brfrn169: Sounds good! Thank you for the explanation!

brfrn169 (approved):
LGTM! Thank you!

docs/how-to-deploy-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+## Deploy ScalarDB Analytics with PostgreSQL
+
+```console
+helm install <RELEASE_NAME> scalar-labs/scalardb-analytics-postgresql -n <NAMESPACE> -f /<PATH_TO_YOUR_CUSTOM_VALUES_FILE_FOR_SCALARDB_ANALYTICS_WITH_POSTGRESQL> --version <CHART_VERSION>
+```
+
+## Upgrade a ScalarDB Analytics with PostgreSQL deployment
+
```
josh-wong, line 20: Similar to my previous comment, I'd like to add a sentence here for context.

```suggestion
## Upgrade a ScalarDB Analytics with PostgreSQL deployment

To upgrade a ScalarDB Analytics with PostgreSQL deployment, run the following command, replacing the contents in the angle brackets as described:

```

docs/how-to-deploy-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,29 @@
...
+```console
+helm upgrade <RELEASE_NAME> scalar-labs/scalardb-analytics-postgresql -n <NAMESPACE> -f /<PATH_TO_YOUR_CUSTOM_VALUES_FILE_FOR_SCALARDB_ANALYTICS_WITH_POSTGRESQL> --version <CHART_VERSION>
+```
+
+## Delete a ScalarDB Analytics with PostgreSQL deployment
+
```
josh-wong, line 26: Similar to my previous comment, I'd like to add a sentence here for context.

```suggestion
## Delete a ScalarDB Analytics with PostgreSQL deployment

To delete a ScalarDB Analytics with PostgreSQL deployment, run the following command, replacing the contents in the angle brackets as described:  

```

josh-wong (changes requested):
Looking good!👍 I've added some comments and suggestions, so PTAL!

docs/configure-custom-values-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,179 @@
+# Configure a custom values file for ScalarDB Analytics with PostgreSQL
+
+This document explains how to create your custom values file for the ScalarDB Analytics with PostgreSQL chart. For details on the parameters, see the [README](https://github.com/scalar-labs/helm-charts/blob/main/charts/scalardb-analytics-postgresql/README.md) of the ScalarDB Analytics with PostgreSQL chart.
+
+## Required configurations
```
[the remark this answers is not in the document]
kota2and3kan: Looks good! Thank you for your suggestion!

kota2and3kan:
@komamitsu @josh-wong 
Thank you for your review!
I applied your suggestions.

charts/scalardb-analytics-postgresql/values.yaml
```diff
@@ -33,31 +33,31 @@ scalardbAnalyticsPostgreSQL:
...
       maxSurge: 25%
-      # -- The number of pods that can be unavailable during the update process
+      # -- The number of pods that can be unavailable during the update process.
       maxUnavailable: 25%
     # -- New pods are added gradually, and old pods are terminated gradually, e.g: Recreate or RollingUpdate
     type: RollingUpdate
 
   service:
-    # -- service types in kubernetes
+    # -- Service types in kubernetes
```
josh-wong, line 60: Sorry, I missed the capitalization of `Kubernetes` here.

```suggestion
    # -- Service types in Kubernetes
```

josh-wong (approved):
I left one minor suggestion for something that I didn't catch before. Other than that, LGTM! Thank you🙇‍♂️

komamitsu (approved):
LGTM! 👍

charts/scalardb-analytics-postgresql/values.yaml
kota2and3kan: Oh, I also overlooked it. Thank you! 🙇‍♀️

charts/scalardb-analytics-postgresql/README.md
choplin: Just curious. How is this file generated?

docs/configure-custom-values-scalardb-analytics-postgresql.md
```diff
@@ -0,0 +1,185 @@
...
+    repository: <SCHEMA_IMPORTER_CONTAINER_IMAGE>
+```
+
+### Replica configurations (optional based on your environment)
+
+You can specify the number of ScalarDB Analytics with PostgreSQL replicas (pods) by using `scalardbAnalyticsPostgreSQL.replicaCount`.
+
+```yaml
+scalardbAnalyticsPostgreSQL:
+  replicaCount: 3
```
choplin, line 141: Does this setting create the Postgres instances with replication? Or does it just make multiple instances?

choplin (approved):
Overall, it looks good to me! Thanks! I left several questions. I would appreciate it if you could take a look.

charts/scalardb-analytics-postgresql/README.md
kota2and3kan: You can generate this file by using `helm-docs`.
https://github.com/norwoodj/helm-docs

Also, in our repository, you can run the `helm-docs` by using the following script.
https://github.com/scalar-labs/helm-charts/blob/main/scripts/update-chart-docs.sh

docs/configure-custom-values-scalardb-analytics-postgresql.md
kota2and3kan: This setting just makes multiple instances without streaming replication or logical replication. At the moment, we cannot control the replication or HA feature of PostgreSQL by using this chart. We just deploy it as a single instance or multiple instances.

As I mentioned, we cannot control HA features, however, I think ScalarDB Analytics with PostgreSQL is basically a `read-only` product for the analytical workload. So, I don't think we need to use the replication feature at this time.

But, from the perspective of availability, I think we can make it more available by deploying 3 pods (ideally across on 3 zones in a cloud environment).

This is why I set `3` by default in this chart.

kota2and3kan (commented):
@choplin 
Thank you for your review and questions!
I left answers. Please take a look when you have time!

PR #249, opened 2024-01-25
Revise scalar-loki-stack-custom-values.yaml to scrape the logs of scalar-admin-for-kubernetes

supl:
## Description

This PR adds a configuration to `scalar-loki-stack-custom-values.yaml` to make Loki to scrape the logs of the pods that run scalar-admin-for-kubernetes. This is an important configuration because the results of scalar-admin-for-kubernetes are displayed as logs.

## Related issues and/or PRs

N/A

## Changes made

- revised `scalar-loki-stack-custom-values.yaml`

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [ ] Tests (unit, integration, etc.) have been added for the changes.
- [ ] Any dependent changes in other PRs have been merged and published.

### DIFF
--- a/docs/conf/scalar-loki-stack-custom-values.yaml
+++ b/docs/conf/scalar-loki-stack-custom-values.yaml
@@ -57,6 +57,7 @@ promtail:
         - job_name: scalar-admin-for-kubernetes
           pipeline_stages:
             - docker: {}
+            - cri: {}
           kubernetes_sd_configs:
             - role: pod
           relabel_configs:
--- a/docs/conf/scalar-loki-stack-custom-values.yaml
+++ b/docs/conf/scalar-loki-stack-custom-values.yaml
@@ -53,3 +53,27 @@ promtail:
                 - __meta_kubernetes_pod_uid
                 - __meta_kubernetes_pod_container_name
               target_label: __path__
+        # -- the `scalar-admin-for-kubernetes` job scrapes all the logs of Scalar Admin for Kubernetes Pods
+        - job_name: scalar-admin-for-kubernetes
+          pipeline_stages:
+            - docker: {}
+          kubernetes_sd_configs:
+            - role: pod
+          relabel_configs:
+            - source_labels:
+                - __meta_kubernetes_pod_node_name
+              target_label: __host__
+            - action: replace
+              source_labels:
+                - __meta_kubernetes_pod_name
+              target_label: pod
+            - action: keep
+              regex: (.*)scalar-admin-for-kubernetes-(.+)
+              source_labels:
+                - pod
+            - replacement: /var/log/pods/*$1/*.log
+              separator: /
+              source_labels:
+                - __meta_kubernetes_pod_uid
+                - __meta_kubernetes_pod_container_name
+              target_label: __path__


docs/conf/scalar-loki-stack-custom-values.yaml
```diff
@@ -53,3 +53,27 @@ promtail:
                 - __meta_kubernetes_pod_uid
                 - __meta_kubernetes_pod_container_name
               target_label: __path__
+        # -- the `scalar-admin-for-kubernetes` job scrapes all the logs of Scalar Admin for Kubernetes Pods
```
feeblefakie, line 56: Sorry that I'm not familiar with the Loki configurations, and it might not be directly related to this PR; let me ask several questions.

- Do we need to scrape all the logs in every job execution?
- scraped results are stored anywhere persistent?  Or if Loki logs are removed or rotated, are they gone?

Anyway, we should keep the paused duration persistently independent from the Loki logs lifecycle, and I'm wondering if we can achieve that.

docs/conf/scalar-loki-stack-custom-values.yaml
```diff
@@ -53,3 +53,28 @@ promtail:
...
+        - job_name: scalar-admin-for-kubernetes
+          pipeline_stages:
+            - docker: {}
+            - cri: {}
```
supl, line 60: This is necessary since many managed Kubernetes cluster is using CRI but not Docker.

kota2and3kan: This is not related to this PR directly, but do we need to add this `cri: {}` configuration in both ScalarDB and ScalarDL configurations?

kota2and3kan (commented):
Sorry for my late review... 🙇‍♀️
I was able to see the log of Scalar Admin for Kubernetes on the Grafana dashboard by using this configuration.
So, overall looks good to me, but I have one question.

supl:
@kota2and3kan 

Sorry for my late response.
I nearly forgot about this PR.

Thank you for the reviewing. Regarding your question,

> This is not related to this PR directly, but do we need to add this cri: {} configuration in both ScalarDB and ScalarDL configurations?

I think we can do it as well because I guess we only need the `message` part.
https://grafana.com/docs/loki/latest/send-data/promtail/stages/cri/

The `docker: {}` pipe did this for me so in the `minikube + docker` case we only see the `message` part of the Docker log format.
https://grafana.com/docs/loki/latest/send-data/promtail/stages/docker/

What do you think?

kota2and3kan:
> > This is not related to this PR directly, but do we need to add this cri: {} configuration in both ScalarDB and ScalarDL configurations?
> 
> I think we can do it as well because I guess we only need the `message` part. https://grafana.com/docs/loki/latest/send-data/promtail/stages/cri/
> 
> The `docker: {}` pipe did this for us so in the `minikube + docker` case we only see the `message` part of the Docker log format. https://grafana.com/docs/loki/latest/send-data/promtail/stages/docker/
> 
> What do you think?

@supl 
Thank you for the discussion in the other meeting!

I understood the behavior of configuration of `docker: {}` and `cri: {}`.

As we discussed in the meeting, let's keep this PR as is and update configurations for ScalarDB and ScalarDL as another task after we test!

kota2and3kan (approved):
LGTM! Thank you!

supl:
@kota2and3kan 

Thank you.
I created an issue for the deprecation of `docker` pipe.
https://github.com/scalar-labs/helm-charts/issues/250

FYI.

docs/conf/scalar-loki-stack-custom-values.yaml
supl: @feeblefakie 

I had a discussion with Yokoyama-san and confirm the feasibility.
In short, Scalar Manager will periodically (maybe per minute) get paused durations from Loki and preserve them in the ConfigMap. So, it doesn't matter how users configure Loki's retention or storage. Scalar Manager provides paused duration by itself.

feeblefakie: @supl So, the answers to the questions are both No?

feeblefakie: If the answer to the first question is yes, it might be an issue. For example, if the job scrapes all the logs in every execution, and users configure a long retention period, I think it might make the job execution take quite long and expensive.

supl: @feeblefakie 

Sorry for my late response.

> Do we need to scrape all the logs in every job execution?

In the comment `the `scalar-admin-for-kubernetes` job scrapes all the logs of Scalar Admin for Kubernetes Pods`, the `all` is for pods.

Maybe it's better to changed the sentence to `the `scalar-admin-for-kubernetes` job scrapes the logs of all Scalar Admin for Kubernetes Pods`

The Promtail job is a progressive process (tail) that keeps shipping new updated logs to Loki. So it's no problem even it has been running and shipping many logs for a very long time.

> scraped results are stored anywhere persistent? Or if Loki logs are removed or rotated, are they gone?

1. Loki has no retention by default unless users explicitly configure it. If users specify a retention, the Loki remove the logs older than the retention from its storage.
2. Loki's default storage is file-based similar to `/var/logs`. If we don't use a permanent volume, or change the storage to a cloud service, for example, DynamoDB, we will loss the logs when Loki server is restarted.

feeblefakie (approved):
LGTM!

PR #254, opened 2024-02-27
[scalar-manager] Create the Helm chart for Scalar Manager

supl:
## Description

This PR deletes the previous Helm chart of Scalar Manager and creates a new one from scratch.

This is my first trial. Please forgive me if I missed some practice and didn't do it properly.

## Related issues and/or PRs

N/A

## Changes made

- added templates
- added values.yaml
- added Chart.yaml
- added other related files

## Checklist


charts/scalar-manager/Chart.yaml
```diff
@@ -5,13 +5,17 @@ type: application
 version: 2.0.0-SNAPSHOT
 appVersion: 2.0.0-SNAPSHOT
```
supl, line 6: Now we only have 1.0.0-SNAPSHOT images.
I use 2.0.0-SNAPSHOT here because we will bump the SNAPSHOT images version to 2.0.0 after officially release Scalar Manager.

charts/scalar-manager/templates/scalar-manager/clusterrole.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "scalar-manager.fullname" . }}
+  labels:
+    {{- include "scalar-manager.labels" . | nindent 4 }}
+rules:
+  - apiGroups: ["", "apps", "batch", "rbac.authorization.k8s.io"]
+    resources: ["pods", "deployments", "services", "namespaces", "configmaps", "secrets", "cronjobs", "serviceaccounts", "roles", "rolebindings", "jobs"]
+    verbs: ["get", "list", "create", "patch", "delete", "update"]
```
supl, line 10: I put all resources and verbs in one rule for convenience.
However, this rule grants too many permissions.
I guess in better practice we should separate them accordingly?

superbrothers: In general, you should define separate rules for each api group. For example, the current configuration grants permissions to a pod resource in the `apps` group that does not exist. This can be a problem if this resource is created in the future and should not be granted.

supl: Thank you for your advise.
I made a change according to your suggestion in https://github.com/scalar-labs/helm-charts/pull/254/commits/31825192575f7b16e40e81af516a6dc1369b2e56

The rules became to

```yaml
rules:
  - apiGroups: [""]
    resources: ["pods", "services", "namespaces", "configmaps", "secrets", "serviceaccounts"]
    verbs: ["get", "list", "create", "patch", "delete", "update"]
  - apiGroups: ["batch"]
    resources: ["cronjobs", "jobs"]
    verbs: ["get", "list", "create", "delete"]
  - apiGroups: ["apps"]
    resources: ["deployments"]
    verbs: ["get", "list"]
  - apiGroups: ["rbac.authorization.k8s.io"]
    resources: ["roles", "rolebindings"]
    verbs: ["get", "list", "create", "delete"]
```

In the core group, for example, this application doesn't delete or update pods. If so, should we separate the resources with proper verbs in general practice?

Like, for example,

```yaml
rules:
  - apiGroups: [""]
    resources: ["configmaps", "secrets", "serviceaccounts"]
    verbs: ["get", "list", "create", "patch", "delete", "update"]
  - apiGroups: [""]
    resources: ["pods", "services", "namespaces"]
    verbs: ["get", "list"]
```

kota2and3kan (commented):
Thank you for updating the chart!
I left some suggestions and questions. 

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+    {{- include "scalar-manager.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "scalar-manager.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/scalar-manager/configmap.yaml") . | sha256sum }}
```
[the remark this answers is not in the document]
kota2and3kan: As we discussed in another comment, we decided to use `application.properties`. So, we will keep this annotation to apply the updated `application.properties` properly when we run the `helm upgrade` command.

charts/scalar-manager/templates/scalar-manager/configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
kota2and3kan: As we discussed, we noticed that this `ConfigMap` overwrites some stored metadata (e.g., managed cluster or pause duration info) when we run the `helm upgrade` command. So, it would be better to remove this `ConfigMap` from the helm chart side and create it on the Scalar Manager API side.

supl:
@kota2and3kan 

Thank you for your advice.

I made the change according to your suggestion in https://github.com/scalar-labs/helm-charts/pull/254/commits/aeb8ee4267a46477e90196b78e3f42e6e469bb96

I created an issue #269 for the ClusterRole issue we discussed earlier

charts/scalar-manager/templates/scalar-manager/clusterrole.yaml
superbrothers: > this application doesn't delete or update pods. If so, should we separate the resources with proper verbs in general practice?

Yes, that is correct. Based on the principle of least privilege, unnecessary privileges should not be granted.

supl: @superbrothers 
Thank you very much.

Sorry, I recalled that we probably need these permissions because the application in this Helm chart (Scalar Manager) uses Helm to install another Helm chart. Allow me to keep them.

feeblefakie (approved):
LGTM!

charts/scalar-manager/templates/scalar-manager/configmap.yaml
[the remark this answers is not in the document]
kota2and3kan: We address this issue on the Scalar Manager API side.
https://github.com/scalar-labs/scalar-manager-api/pull/90

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -0,0 +1,74 @@
...
+              subPath: scalar-manager-api-application.properties
+        - name: {{ .Chart.Name }}-web
+          image: "{{ .Values.web.image.repository }}:{{ .Values.web.image.tag | default .Chart.AppVersion }}"
+          resources:
+            {{- toYaml .Values.web.resources | nindent 12 }}
+          ports:
+            - containerPort: 3000
+          imagePullPolicy: {{ .Values.web.image.pullPolicy }}
+          securityContext:
+            runAsUser: 1000
```
kota2and3kan, line 51: ## Question 1

```suggestion
            runAsUser: 1001
```
It seems that we use `1001` on the `Dockerfile` side of Scalar Manager API.
https://github.com/scalar-labs/scalar-manager-web/blob/main/Dockerfile#L41

Is this UID `1000` correct? Or, we should use `1001` instead of `1000`?

## Question 2

It seems that we specify the `USER` directive on the `Dockerfile` side, and we set `runAsNonRoot: true` on the helm chart side as a default value.

In this case, I think the `scalar-manager-api` container image will run with UID `1001` automatically.

Do we really need to specify `runAsUser` as a hard-coded value here? If we don't need it, I think it would be better to remove this hard-coded value from here. (For example, I guess this hard-coded value might cause some issues in the OpenShift environment.)

kota2and3kan (commented):
@supl 
Thank you for the updates!
As I mentioned before, I updated minor things in 819baf3abac72818d194c2de1a2a9ec10de99392 and 590d91191db1178bf706ae43ae3f9794e4303563.

Also, I have two questions about the `runAsUser` configuration.
Could you please take a look when you have time? 🙇‍♀️

charts/scalar-manager/templates/scalar-manager/deployment.yaml
supl: @kota2and3kan 
Thank you for your advice.

1000 was a typo. It should be `1001`.

Regarding the question 2.

I did this because I encountered an issue from creating the container with the `runAsNonRoot` security context. This is the error message from `kubectl describe pod <pod-name>`.

```
  Warning  Failed     12s (x7 over 65s)  kubelet            Error: container has runAsNonRoot and image has non-numeric user (nextjs), cannot verify user is non-root (pod: "scalar-manager-57cdcb57d4-pbvvc_scalar-manager(da3940c6-0d80-4287-aa2a-1e9b1f8adc6d)", container: scalar-manager-web)
```

I am not sure if it's because of the USER command in the Dockerfile here https://github.com/scalar-labs/scalar-manager-web/blob/main/Dockerfile#L54 

It specifies the user name but not the user ID.

I tried to change it locally with
```
USER 1001
```

like we did for the API
https://github.com/scalar-labs/scalar-manager-api/blob/main/Dockerfile#L22

However, my internet is too slow here to finish building the image so I couldn't really test it.
(the node packages are just a mess ...

Anyway, since it's not a good practice to use `runAsUser` context, I removed it in https://github.com/scalar-labs/helm-charts/pull/254/commits/5502ca00d60537a0c9e4eb10b6e81c393b47ffb3

I will create a PR in the web repository to specify the user by user ID instead of the user name and try to test it later with better Internet conditions.

supl: The PR using the numeric user ID instead of the user name is submitted in
https://github.com/scalar-labs/scalar-manager-web/pull/70

supl: I guess the CI failed because of the removal of `runAsUser`.
We can test it after we merge https://github.com/scalar-labs/scalar-manager-web/pull/70 (the workflow will create the snapshot image and we can re-run the action here)

kota2and3kan: @supl 
Thank you for the updates!

As I mentioned on the Scalar Manager Web side, according to the code comment of Kubernetes, it seems that we MUST use `number (UID)` instead of `name (string)` in the container image if we want to use a non-root user.

- https://github.com/kubernetes/kubernetes/blob/v1.30.1/pkg/kubelet/kuberuntime/kuberuntime_container.go#L331
- https://github.com/kubernetes/kubernetes/blob/v1.30.1/pkg/kubelet/kuberuntime/security_context_others.go#L48-L49

kota2and3kan:
Thank you for the updates!

Overall looks good to me, but I am waiting for the release of Scalar Manager Web and Scalar Manager API, just in case.

After they are released (and the CI passed), I will merge this PR and release the helm chart.

supl:
@kota2and3kan 
As we discussed earlier, I changed the snapshot version to `3.0.0` for the main branch in https://github.com/scalar-labs/helm-charts/pull/254/commits/b8864ebd7b5be8f002dfe6289d2182e4ecf04c36

kota2and3kan (approved):
LGTM! Thank you!

PR #262, opened 2024-04-26
[envoy] Support cert-manager in Scalar Envoy chart

kota2and3kan:
## Description

This PR updates Scalar Envoy chart to support cert-manager.

Please take a look!

## Related issues and/or PRs

- https://github.com/scalar-labs/scalar-envoy/pull/30
  - I updated Scalar Envoy to support cert-manager
- https://github.com/scalar-labs/docs-internal-orchestration/pull/24
  - You can see the documents that describe how to use these new features in this PR.

## Changes made

Mainly, this PR adds/updates the following features (manifests):

1. Deploy a `Certificate` resource for cert-manager by using `charts/envoy/templates/certmanager.yaml`.

2. Generate self-signed certificate by using self-signed CA if you don't specify any `Issuer` resources in `issuerRef` configuration.
   - If you deploy Scalar Envoy as a sub chart with upstream charts (ScalarDB Cluster, ScalarDL Ledger, or ScalarDL Auditor charts) without `issuerRef` configuration, upstream charts deploy self-signed root CA. And, this Envoy chart will use it to create certificate for Envoy.

   ```
   [Upstream charts] ---(Deploy)---> [Issuer (self-signed CA)] <---(Refer via issuerRef configuration)--- [Certificate resource] <---(Deploy)--- [Envoy chart]
   ```

3. Update `deployment.spec.template.spec.volumes` and `deployment.spec.template.spec.containers.volumeMounts` to remove `subPath` configuration.
   - Cert-manager creates a secret resource that includes private key and certificate and we use it by mounting it on pods. And, cert-manager automatically renew (update) the certificate in the secret resource before the certificate will expired. However, if we use `subPath` to specify the mounted file name, [Kubernetes does not apply the changes of the secret resource](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod). So, to take advantage of cert-manager's automatically renew feature, we should not use `subPath`. This is because I removed `subPath` from the `volumeMounts` configuration.

4. Update file names of private key and certificate.
   - From the perspective of cert-manager specification (strictly, [TLS Secrets](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) specification of Kubernetes), we must use the fixed names for private key and certificate files. This is because cert-manager creates a secret resource that includes private key and certificate files with fixed names based on the [TLS Secrets](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) specification. So, I updated the file names in existing manifests.
     - `cert.pem` -> `tls.crt`
     - `key.pem` -> `tls.key`
     - `ca.pem` -> `ca.crt`

## Checklist

N/A

## Release notes

Support cert-manager in Scalar Envoy chart. You can manage private key and certificate for TLS connections by using cert-manager.


charts/envoy/templates/deployment.yaml
```diff
@@ -82,36 +82,37 @@ spec:
...
-        {{- if .Values.tls.upstream.caRootCertSecret }}
-        - name: scalar-envoy-tls-ca-root-volume
+        {{- if .Values.tls.upstream.enabled }}
+        - name: scalar-envoy-tls-upstream-volume
           secret:
+            {{- if .Values.tls.upstream.caRootCertSecret }}
             secretName: {{ .Values.tls.upstream.caRootCertSecret }}
+            {{- else }}
+            secretName: {{ include "envoy.fullname" . }}-envoy-tls-cert
+            {{- end }}
```
kota2and3kan, line 101: If you set `tls.upstream.caRootCertSecret` explicitly, this chart uses the specified root CA certificate. However, if you don't specify that parameter, this chart uses the root CA certificate for downstream TLS.

In other words, if you use the same CA to issue certificate for Envoy and upstream Scalar products, you don't need to set root CA certificate manually.

You can see the conditions on the document side.
https://github.com/scalar-labs/docs-internal-orchestration/pull/24/files#r1584029281

superbrothers (commented):
envoy sub chart does not seem to be configured to use a CA certificate for communication with the ledger. Does the user need to specify this explicitly with chart alues?

charts/envoy/values.yaml
```diff
@@ -124,6 +124,29 @@ tls:
     certChainSecret: ""
     # -- Name of the Secret containing the private key file used for TLS communication.
     privateKeySecret: ""
+    certManager:
+      # -- Use cert-manager to manage private key and certificate files.
+      enabled: false
+      # -- Duration of a certificate.
+      duration: "87600h0m0s"
```
feeblefakie, line 131: Isn't it 3650 days? Should it be `8760h`?

feeblefakie (commented):
Overall, looking good. Left one comment. PTAL!

charts/envoy/values.yaml
kota2and3kan: Ah, yes. You are right.
According to the previous offline discussion, I will update this default value to `8760h (365 days)`.

kota2and3kan: Fixed in 26dc8459235933e38de6bcdae4e84c1909d9b1ce.

PR #264, opened 2024-04-26
[scalardl-ledger] Support cert-manager in ScalarDL Ledger chart

kota2and3kan:
## Description

This PR updates ScalarDL Ledger chart to support cert-manager.

Please take a look!

## Related issues and/or PRs

- https://github.com/scalar-labs/docs-internal-orchestration/pull/24
  - You can see the documents that describe how to use these new features in this PR.
- https://github.com/scalar-labs/helm-charts/pull/262
  - If you use self-signed CA, this chart deploy self-signed CA. And, Scalar Envoy chart refer it to generate certificate for Envoy.

## Changes made

Mainly, this PR adds/updates the following features (manifests):

1. Deploy a `Certificate` resource for cert-manager by using `charts/scalardl/templates/ledger/certmanager.yaml`.

2. Deploy self-signed CA and create certificate for ScalarDL Ledger if you don't specify any `Issuer` resources in `issuerRef` configuration.

   - If you deploy Scalar Envoy as a sub chart of this ScalarDL Ledger chart without `issuerRef` configuration, the Envoy chart will use self-signed CA that is deployed by this chart to create certificate for Envoy.

   ```
   [ScalarDL Ledger chart (this chart)] ---(Deploy)---> [Issuer (self-signed CA)] <---(Refer via issuerRef configuration)--- [Certificate resource] <---(Deploy)--- [Envoy chart]
   ```

3. Update `deployment.spec.template.spec.volumes` and `deployment.spec.template.spec.containers.volumeMounts` to remove `subPath` configuration.

   - Cert-manager creates a secret resource that includes private key and certificate, and we use them by mounting them on pods. And, cert-manager automatically renew (update) the certificate in the secret resource before the certificate will be expired. However, if we use `subPath` to specify the mounted file name, [Kubernetes does not apply the changes of the secret resource](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod). So, to take advantage of cert-manager's automatically renew feature, we should not use `subPath`. This is because I removed `subPath` from the `volumeMounts` configuration.

4. Update file names of private key and certificate.

   - From the perspective of cert-manager specification (strictly, [TLS Secrets](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) specification of Kubernetes), we must use the fixed names for private key and certificate files. This is because cert-manager creates a secret resource that includes private key and certificate files with fixed names based on the [TLS Secrets](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) specification. So, I updated the file names in existing manifests.
     - `cert.pem` -> `tls.crt`
     - `key.pem` -> `tls.key`
     - `ca.pem` -> `ca.crt`

## Checklist

N/A

## Release notes

Support cert-manager in ScalarDL Ledger chart. You can manage private key and certificate for TLS connections by using cert-manager.

### DIFF
--- a/charts/scalardl/README.md
+++ b/charts/scalardl/README.md
@@ -77,12 +77,12 @@ Current chart version is `5.0.0-SNAPSHOT`
 | ledger.tls.caRootCertSecret | string | `""` | Name of the Secret containing the custom CA root certificate for TLS communication. |
 | ledger.tls.certChainSecret | string | `""` | Name of the Secret containing the certificate chain file used for TLS communication. |
 | ledger.tls.certManager.dnsNames | list | `["localhost"]` | Subject Alternative Name (SAN) of a certificate. |
-| ledger.tls.certManager.duration | string | `"87600h0m0s"` | Duration of a certificate. |
+| ledger.tls.certManager.duration | string | `"8760h0m0s"` | Duration of a certificate. |
 | ledger.tls.certManager.enabled | bool | `false` | Use cert-manager to manage private key and certificate files. |
 | ledger.tls.certManager.issuerRef | object | `{}` | Issuer references of cert-manager. |
 | ledger.tls.certManager.privateKey | object | `{"algorithm":"ECDSA","encoding":"PKCS1","size":256}` | Configuration of a private key. |
 | ledger.tls.certManager.renewBefore | string | `"360h0m0s"` | How long before expiry a certificate should be renewed. |
-| ledger.tls.certManager.selfSignedCaRootCert | object | `{"duration":"87600h0m0s","renewBefore":"360h0m0s"}` | Configuration of a certificate for self-signed CA. |
+| ledger.tls.certManager.selfSignedCaRootCert | object | `{"duration":"8760h0m0s","renewBefore":"360h0m0s"}` | Configuration of a certificate for self-signed CA. |
 | ledger.tls.certManager.usages | list | `["server auth","key encipherment","signing"]` | List of key usages. |
 | ledger.tls.enabled | bool | `false` | Enable TLS. You need to enable TLS when you use wire encryption feature of ScalarDL Ledger. |
 | ledger.tls.overrideAuthority | string | `""` | The custom authority for TLS communication. This doesn't change what host is actually connected. This is intended for testing, but may safely be used outside of tests as an alternative to DNS overrides. For example, you can specify the hostname presented in the certificate chain file that you set by using `ledger.tls.certChainSecret`. This chart uses this value for startupProbe and livenessProbe. |
--- a/charts/scalardl/README.md
+++ b/charts/scalardl/README.md
@@ -76,6 +76,14 @@ Current chart version is `5.0.0-SNAPSHOT`
 | ledger.strategy.type | string | `"RollingUpdate"` | New pods are added gradually, and old pods are terminated gradually, e.g: Recreate or RollingUpdate |
 | ledger.tls.caRootCertSecret | string | `""` | Name of the Secret containing the custom CA root certificate for TLS communication. |
 | ledger.tls.certChainSecret | string | `""` | Name of the Secret containing the certificate chain file used for TLS communication. |
+| ledger.tls.certManager.dnsNames | list | `["localhost"]` | Subject Alternative Name (SAN) of a certificate. |
+| ledger.tls.certManager.duration | string | `"87600h0m0s"` | Duration of a certificate. |
+| ledger.tls.certManager.enabled | bool | `false` | Use cert-manager to manage private key and certificate files. |
+| ledger.tls.certManager.issuerRef | object | `{}` | Issuer references of cert-manager. |
+| ledger.tls.certManager.privateKey | object | `{"algorithm":"ECDSA","encoding":"PKCS1","size":256}` | Configuration of a private key. |
+| ledger.tls.certManager.renewBefore | string | `"360h0m0s"` | How long before expiry a certificate should be renewed. |
+| ledger.tls.certManager.selfSignedCaRootCert | object | `{"duration":"87600h0m0s","renewBefore":"360h0m0s"}` | Configuration of a certificate for self-signed CA. |
+| ledger.tls.certManager.usages | list | `["server auth","key encipherment","signing"]` | List of key usages. |
 | ledger.tls.enabled | bool | `false` | Enable TLS. You need to enable TLS when you use wire encryption feature of ScalarDL Ledger. |
 | ledger.tls.overrideAuthority | string | `""` | The custom authority for TLS communication. This doesn't change what host is actually connected. This is intended for testing, but may safely be used outside of tests as an alternative to DNS overrides. For example, you can specify the hostname presented in the certificate chain file that you set by using `ledger.tls.certChainSecret`. This chart uses this value for startupProbe and livenessProbe. |
 | ledger.tls.privateKeySecret | string | `""` | Name of the Secret containing the private key file used for TLS communication. |
--- a/charts/scalardl/README.md
+++ b/charts/scalardl/README.md
@@ -82,7 +82,9 @@ Current chart version is `5.0.0-SNAPSHOT`
 | ledger.tls.certManager.issuerRef | object | `{}` | Issuer references of cert-manager. |
 | ledger.tls.certManager.privateKey | object | `{"algorithm":"ECDSA","encoding":"PKCS1","size":256}` | Configuration of a private key. |
 | ledger.tls.certManager.renewBefore | string | `"360h0m0s"` | How long before expiry a certificate should be renewed. |
-| ledger.tls.certManager.selfSignedCaRootCert | object | `{"duration":"8760h0m0s","renewBefore":"360h0m0s"}` | Configuration of a certificate for self-signed CA. |
+| ledger.tls.certManager.selfSigned.caRootCert.duration | string | `"8760h0m0s"` | Duration of a self-signed CA certificate. |
+| ledger.tls.certManager.selfSigned.caRootCert.renewBefore | string | `"360h0m0s"` | How long before expiry a self-signed CA certificate should be renewed. |
+| ledger.tls.certManager.selfSigned.enabled | bool | `false` | Use self-signed CA. |
 | ledger.tls.certManager.usages | list | `["server auth","key encipherment","signing"]` | List of key usages. |
 | ledger.tls.enabled | bool | `false` | Enable TLS. You need to enable TLS when you use wire encryption feature of ScalarDL Ledger. |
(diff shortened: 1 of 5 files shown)


charts/scalardl/templates/ledger/deployment.yaml
```diff
@@ -43,20 +43,21 @@ spec:
...
+        - name: scalardl-ledger-tls-volume
+          projected:
+            sources:
+              - secret:
+                  name: {{ .Values.ledger.tls.caRootCertSecret }}
+              - secret:
+                  name: {{ .Values.ledger.tls.certChainSecret }}
+              - secret:
+                  name: {{ .Values.ledger.tls.privateKeySecret }}
       {{- end }}
```
kota2and3kan, line 61: If you don't use cert-manager (and you want to manage each file separately), you have to create three secrets for the private key, certificate, and CA certificate.

In this case, I want to combine them into one volume to mount them easily under the same directory in the container. So, I use `projected` volume here.

charts/scalardl/templates/ledger/deployment.yaml
```diff
@@ -146,9 +136,7 @@ spec:
               - -addr=localhost:50051
               {{- if .Values.ledger.tls.enabled }}
               - -tls
-              {{- if .Values.ledger.tls.caRootCertSecret }}
-              - -tls-ca-cert=/tls/certs/ca-root-cert.pem
-              {{- end }}
+              - -tls-ca-cert=/tls/scalardl-ledger/certs/ca.crt
```
kota2and3kan, line 139: When you enable TLS, the `-tls-ca-cert` option is mandatory. In other words, `{{- if .Values.scalardbCluster.tls.caRootCertSecret }}` does not make sense. So, I removed that unnecessary condition.

charts/scalardl/values.yaml
```diff
@@ -245,3 +245,29 @@ ledger:
...
+        encoding: PKCS1
+        size: 256
+      # -- List of key usages.
+      usages:
+        - server auth
+        - key encipherment
+        - signing
+      # -- Subject Alternative Name (SAN) of a certificate.
+      dnsNames:
+        - localhost
```
kota2and3kan, line 271: I set `localhost` as a default value of the `dnsNames`. This is because:

- The `grpc_health_probe` command accesses ScalarDL Ledger by specifying `localhost` like `-addr=localhost:50051`.
- The `grpc_health_probe` command verifies the certificate by using `Subject Alternative Name (SAN)` instead of `Common Name (CN)` of the certificate. In other words, at least one SAN configuration is necessary. (I will update the document in another PR.)

supl (approved):
LGTM.
Thank you.

I have a general question about the self signing certificate chain.

In our practice, we will build two `cert-manager.io/v1/Issuer` for
- *-self-signed-issuer
- *-ca-issuer

The `*-self-signed-issuer` seems only used to issue the certificate to `*-ca-issuer`. 

I am not sure whether the reason for doing that is because the CA Issuer resource needs to be created with `ca` specification.

I realize the Issuer need to create with 
```
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: {{ include "scalardl.fullname" . }}-ca-issuer
  namespace: {{ .Release.Namespace }}
spec:
  ca:
    secretName: {{ include "scalardl.fullname" . }}-root-ca-cert
```

So we need the very first Certificate (issued by the `*-self-signed-issuer`) before we create the CA Issuer

I don't know if it's possible. I am curious if the CA can sign the root certificate by itself?

The

```
  issuerRef:
    name: {{ include "scalardl.fullname" . }}-self-signed-issuer
```

becomes

```
  issuerRef:
    name: {{ include "scalardl.fullname" . }}-ca-issuer
```

I guess it's difficult...

The reason I have this question is because the `root` certificates are commonly self-signed so I feel it makes the term `root` more close to the real world if we can remove the role `*-self-signer-issuer` in our case. (Or maybe we can remove the `root` naming in the Certificate resource https://github.com/scalar-labs/helm-charts/pull/264/files#diff-5f776de94b8a86f1c31686ec1f8895b0e9008479acff64da19dc460eb6bce00dR18 to avoid the confusion.

Sorry, maybe my understanding is not correct...

kota2and3kan:
@supl  
Thank you for your review and question!

> I don't know if it's possible. I am curious if the CA can sign the root certificate by itself?

Sorry, I'm not sure whether we can do it or not...

However, I created those two `Issuer` resource manifests for self-signed CA by following the official document of cert-manger.
https://cert-manager.io/docs/configuration/selfsigned/

And, it seems that `kube-prometheus-stack` adopts the same way, which is described in the above cert-manager documents. (In other words, this way might be a general/popular way to generate a self-signed certificate by using cert-manager.)
https://github.com/prometheus-community/helm-charts/blob/kube-prometheus-stack-58.2.2/charts/kube-prometheus-stack/templates/prometheus-operator/certmanager.yaml

Also, I don't have any clear motivations to adopt a different way from the cert-manager's official document at this point. (Vice versa, I think there might be some risks something like unexpected errors by adopting a different way from the cert-manager's official document.)

These are why I adopt the current way (there are two `Issuer` resources `*-self-signed-issuer` and `*-ca-issuer`) in this chart.

supl:
@kota2and3kan 

Thank you for the explanation!
Understood. There's no problem.
I am just curious if it's even possible 😄

kota2and3kan:
@supl  
Thank you! I keep the current implementation as is!

Your concern makes sense to me. Because when I saw the official document first, I was also very confused... But, this chart deploys all self-signed CA related resources under the abstraction by Helm Chart.   So, I guess users can use it with less confusion than using cert-manager directly by themselves (Although they might be confused even if this implementation...).

charts/scalardl/README.md
```diff
@@ -76,6 +76,14 @@ Current chart version is `5.0.0-SNAPSHOT`
 | ledger.strategy.type | string | `"RollingUpdate"` | New pods are added gradually, and old pods are terminated gradually, e.g: Recreate or RollingUpdate |
 | ledger.tls.caRootCertSecret | string | `""` | Name of the Secret containing the custom CA root certificate for TLS communication. |
 | ledger.tls.certChainSecret | string | `""` | Name of the Secret containing the certificate chain file used for TLS communication. |
+| ledger.tls.certManager.dnsNames | list | `["localhost"]` | Subject Alternative Name (SAN) of a certificate. |
+| ledger.tls.certManager.duration | string | `"87600h0m0s"` | Duration of a certificate. |
```
superbrothers, line 80: It is set to a value greater than the default value, where does this value come from? If ledger supports hot reloading of certificates, then a shorter duration is better.

superbrothers (commented):
envoy sub chart does not seem to be configured to use a CA certificate for communication with the ledger. Does the user need to specify this explicitly with chart values?

charts/scalardl/templates/ledger/certmanager.yaml
```diff
@@ -0,0 +1,83 @@
+{{- if .Values.ledger.tls.certManager.enabled }}
+{{- if not .Values.ledger.tls.certManager.issuerRef }}
```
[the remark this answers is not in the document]
kota2and3kan: > Is there a request from a user who wants to use cert-manager while using an arbitrary issuer?

Yes.

I interviewed the business team about some existing customers' and potential customers' requirements.

As a result, there are some users who want to use the cert-manager with an arbitrary issuer. (i.e., There are some users who don't want to use self-signed certificate for some security reasons.)

> If not, I feel that it would be better to use selfsigned when using the cert-manager function of this Helm Chart, and not use it at all when using an arbitrary certificate.

Ah, I see. Thank you for your suggestion.

I thought there are some users who want to use self-signed certificate. For example, users who have no problem with self-signed certificate or developers who want to test TLS features.

This is a reason way I remained the `self-signed CA` pattern.

However, if they want to use self-signed CA, it is easy to deploy self-signed CA (self-signed issuer) follow the [official docuemnt of cert-manager](https://cert-manager.io/docs/configuration/selfsigned/).

So, now, I'm thinking about it's fine to remove self-signed CA pattern.

Do you have any concerns when we support both `use automatically deployed self-signed CA (self-singed issuer)` and `use arbitrary issuer pattern`?

charts/scalardl/README.md
kota2and3kan: Ah, sorry. I set the default values based on our test environment (because it's a test environment, the duration value is very long.)

However, I didn't confirm the ScalarDL support hot reloading of certificates. So, first, I will ask the development team whether ScalarDL supports hot reloading or not.

If ScalarDL supports hot reloading, I will set the default value to empty (i.e., we will use the default value of `Certificate` spec).

If ScalarDL does not support hot reloading, I will discuss what value is better as a default value again.

kota2and3kan (commented):
@superbrothers 
Thank you for your review and comments!
I left answers in each comment.

---

> envoy sub chart does not seem to be configured to use a CA certificate for communication with the ledger. Does the user need to specify this explicitly with chart values?

Basically, there are 4 patterns in this chart.

1. Use self-signed CA (Use automatically deployed self-signed `Issuer`)
2. Use one arbitrary `Issuer`.
3. Use two different arbitrary `Issuer` for Envoy and Ledger respectively.
4. Not use cert-manager.

In the case of `1.` and `2.`, this chart and envoy subchart mount a `root CA certificate for Ledger` on Envoy pods automatically.

In the case of `3.` or `4.`, users must mount a root CA certificate manually. First, create a `Secret` resource that includes a root CA certificate. Second, set the `Secret` resource name to `envoy.tls.upstream.caRootCertSecret` in the custom values file of Ledger.

You can see the description of this manual step in the following documents.
- https://github.com/scalar-labs/docs-internal-orchestration/blob/2b89161b14f3b6463784846b03db863043f6dce1/docs/en-us/helm-charts/configure-custom-values-envoy.md#enable-tls-in-upstream-connections
- https://github.com/scalar-labs/docs-internal-orchestration/blob/2b89161b14f3b6463784846b03db863043f6dce1/docs/en-us/helm-charts/configure-custom-values-envoy.md#set-your-root-ca-certificate-file-of-upstream-scalar-products

superbrothers (approved):

charts/scalardl/README.md
kota2and3kan: @superbrothers  
FYI:

ScalarDL does not support hot reloading. So, we decided to set `8760h (365 days)` as a default duration.

However, we are planing to support hot reloading on the ScalarDL side. After we implement hot reloading, we will update this default value to `2160h (90 days)`.

jnmt (approved):
LGTM, thank you!

choplin (approved):
LGTM!

feeblefakie (approved):
LGTM! Thank you!

PR #280, opened 2024-12-27
Add auth config for scalar manager

thongdk8:
## Description

This PR adds the configurations in the values for configuring the auth feature in the Scalar Manager.
By default, we can still able to deploy the Scalar Manager without auth enabled.

## Related issues and/or PRs

NA

## Changes made

- Add values for configure Scalar Manger auth
- Add a service for scalar manager API (before we only have service for web)

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

NA

## Release notes

NA


charts/scalar-manager/values.yaml
```diff
@@ -80,12 +84,65 @@ api:
...
     repository: ghcr.io/scalar-labs/scalar-manager-web
     pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
     tag: ""
+
+  authorization:
+    # Whether to enable authorization or not for the web application, if enabled the login, user management page will be available
+    enabled: false
+    # The base URL of the authorization service, default is same as the scalar-manager-api service
+    baseUrl: http://localhost:8080
```
thongdk8, line 140: In the case of the API service running in the same pod with Scalar Manager web doesn't have the persistent endpoints enabled then we can config this `baseUrl` point to another instance of the Scalar Manager API (in other clusters) to do authentication. Note that the JWT public key need to be configured the same in the API instances

charts/scalar-manager/templates/scalar-manager/service.yaml
```diff
@@ -1,16 +1,33 @@
...
     - protocol: TCP
       name: web
-      port: {{ .Values.service.port }}
+      port: {{ .Values.service.web.port }}
       targetPort: 3000
   selector:
     {{- include "scalar-manager.selectorLabels" . | nindent 4 }}
+---
+apiVersion: v1
+kind: Service
```
thongdk8, line 19: Add a service for API in case we want to expose the API for cross Scalar Manager authentication

kota2and3kan (commented):
Thank you for the updates!
I left several suggestions and questions.
Please take a look when you have time! 🙇‍♂️

charts/scalar-manager/templates/scalar-manager/service.yaml
thongdk8: Ah, let me explain it a bit more based on the current spec we have.
The cross authentication is in the case: We have two instances of the SM (Scalar Manager) in different clusters for example, one is deployed with a configured database and it has the authentication endpoints, once login the user will have the JWT tokens to access other protected endpoints from that instance (e.g pause, schedule pause). The other instance is deployed without database config but it is configured with the JWT public key sharing the same as instance 1 so this instance is able to verify requests with the JWT token in the header issued by instance 1, basically client (web UI) can login to SM 1 and access to both SM 1 and SM 2, of course the SM 2 API need to be viable to the SM 1.
Please correct me if I miss something @ypeckstadt 
I made a PR for the docs update that contains a bit more info about deployment configuration here https://github.com/scalar-labs/docs-internal-orchestration/pull/86

charts/scalar-manager/values.yaml
```diff
@@ -80,12 +84,65 @@ api:
...
+
+
 web:
   image:
     repository: ghcr.io/scalar-labs/scalar-manager-web
     pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
     tag: ""
+
+  authorization:
```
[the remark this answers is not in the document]
thongdk8: Actually in the `application.properties` I also used the `authorization` in `app.authorization.enabled` config. The `authentication` word is used in the prefix for the JWT config only `authentication.providers.static-jwt.xxx`.
The main reason that `authorization` makes more sense in our case that the API can be deployed without authentication endpoints but it is still able to verify the request via the JWT in the `Authorization` header from the requests.

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -44,6 +44,13 @@ spec:
           image: "{{ .Values.web.image.repository }}:{{ .Values.web.image.tag | default .Chart.AppVersion }}"
           resources:
             {{- toYaml .Values.web.resources | nindent 12 }}
+          env:
+            - name: NEXT_PUBLIC_AUTH_ENABLED
+              value: {{ .Values.web.authorization.enabled | quote }}
+            - name: NEXT_PUBLIC_PERSISTENCE_API_BASE_URL
+              value: {{ .Values.web.authorization.baseUrl | quote }}
+            - name: NEXT_PUBLIC_OPERATION_API_BASE_URL
+              value: {{ .Values.web.operation.baseUrl | quote }}
```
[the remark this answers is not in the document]
thongdk8: These environment variables are mandatory, but they are for the `web` container. For the `api` container we still keep the config in the `application.properties` as it was.
The config for the web container would not be many I think, for now we will have these new three vars.

charts/scalar-manager/values.yaml
```diff
@@ -27,8 +27,12 @@ nodeSelector: {}
...
 
 service:
-  type: ClusterIP
-  port: 80
+  web:
+    type: ClusterIP
+    port: 80
+  api:
+    type: ClusterIP
+    port: 8080
```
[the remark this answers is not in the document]
thongdk8: Thanks for the suggestion, that makes sense. Let me refactor them.

charts/scalar-manager/values.yaml
```diff
@@ -27,8 +27,12 @@ nodeSelector: {}
 tolerations: []
 
 service:
-  type: ClusterIP
-  port: 80
```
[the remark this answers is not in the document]
thongdk8: Yes, it will break the backward compatibility as the config will be moved under the `web` and I think we will make a new major version release. One other way if we want to keep the backward compatibility is keeping the `service` at the top level as it was for the `web` and add only a `service` config for the API under the `api` config, but that makes the config looks not be in the same style. WDYT?

thongdk8:
@kota2and3kan Thank you for your feedback, I updated the PR, please take a look again when you have time.

charts/scalar-manager/values.yaml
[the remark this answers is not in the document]
kota2and3kan: I see. Thank you for your answer. I think it's fine to release a new major version because there is no actual user now.

charts/scalar-manager/values.yaml
```diff
@@ -80,12 +80,69 @@ api:
...
+    authentication.providers.static-jwt.private-key=${AUTHENTICATION_PROVIDERS_STATIC_JWT_PRIVATE_KEY:}
+    # Public key in X.509/SPKI format using for verifying JWT tokens, when authentication is enabled, this key is used for verifying JWT tokens
+    authentication.providers.static-jwt.public-key=${AUTHENTICATION_PROVIDERS_STATIC_JWT_PUBLIC_KEY:}
+    authentication.providers.static-jwt.issuer-uri=${AUTHENTICATION_PROVIDERS_STATIC_JWT_ISSUER_URI:https://scalar-manager.scalar-labs.com}
+    authentication.providers.static-jwt.access-token-expiration-time=${AUTHENTICATION_PROVIDERS_STATIC_JWT_ACCESS_TOKEN_EXPIRATION_TIME:15m}
+    authentication.providers.static-jwt.refresh-token-expiration-time=${AUTHENTICATION_PROVIDERS_STATIC_JWT_REFRESH_TOKEN_EXPIRATION_TIME:3d}
+
+    # # Initial admin configuration, only need these configurations if persistence endpoints is enabled
+    # app.initial-admin-user.enabled=${APP_INITIAL_ADMIN_USER_ENABLED:false}
+    # app.initial-admin-user.email=${APP_INITIAL_ADMIN_USER_EMAIL:admin@scalar.com}
```
kota2and3kan, line 109:
```suggestion
    # app.initial-admin-user.email=${APP_INITIAL_ADMIN_USER_EMAIL:admin@scalar-labs.com}
```
Is this a typo? I think the domain `scalar\.com` is not our domain.

charts/scalar-manager/values.yaml
```diff
@@ -80,12 +80,69 @@ api:
...
+
+    # Whether to enable authorization or not for the operational endpoints
+    app.authorization.enabled=${APP_AUTHORIZATION_ENABLED:false}
+
+    # JWT configuration
+    # P-256 (secp256k1) private key in PKCS8 format, using for signing JWT tokens, required when persistence endpoints are enabled
+    authentication.providers.static-jwt.private-key=${AUTHENTICATION_PROVIDERS_STATIC_JWT_PRIVATE_KEY:}
+    # Public key in X.509/SPKI format using for verifying JWT tokens, when authentication is enabled, this key is used for verifying JWT tokens
+    authentication.providers.static-jwt.public-key=${AUTHENTICATION_PROVIDERS_STATIC_JWT_PUBLIC_KEY:}
+    authentication.providers.static-jwt.issuer-uri=${AUTHENTICATION_PROVIDERS_STATIC_JWT_ISSUER_URI:https://scalar-manager.scalar-labs.com}
```
kota2and3kan, line 103:
```suggestion
    authentication.providers.static-jwt.issuer-uri=${AUTHENTICATION_PROVIDERS_STATIC_JWT_ISSUER_URI:https://scalar-manager.example.com}
```

Is this URL `https://scalar-manager.scalar-labs.com` provide some actual service? Or, is this just a sample?

If it's a sample, I think it might be better to use another domain like `example.com` that is described in [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html).

charts/scalar-manager/templates/scalar-manager/service.yaml
kota2and3kan: I see. Thank you for explanation!
I have two additional questions.

## Question 1

I understood as follows:

1. Client login to `Scalar Manager 1` that has authentication endpoint.
2. Client get a JWT token from `Scalar Manager 1`.
3. Client can send `pause request` against `Scalar Manager 1` with JWT token.
4. Client can send `pause request` against `Scalar Manager 2` with the same JWT token.

Is my understanding correct?

## Question 2

If the answer of Q1 is yes, which is correct in the following?

- A) Client sends the `pause` request to the endpoint of `Scalar Manager Web`.
- B) Client sends the `pause` request to the endpoint of `Scalar Manager API`.

kota2and3kan (commented):
@thongdk8 
Thank you for your answers!
I left some comments and some additional questions.
Please take a look when you have time! 🙇‍♂️

kota2and3kan:
@thongdk8 
As well as my comment in another PR, I have changed the status of this PR to draft:
https://github.com/scalar-labs/docs-internal-orchestration/pull/86#discussion_r1953520460.

I understand that you are currently working on updating Scalar Manager and will revise this Helm Chart again. (Please correct me if I’m wrong. 🙇‍♂️)

Once you have finished updating Scalar Manager and this Helm Chart, please change the status of this PR back to `ready for review`.

thongdk8:
@kota2and3kan Thank you for checking. As we revised the implementation of Scalar Manager so I will close this PR and make a new one.

PR #303, opened 2025-07-04
Add TLS support for scalar manager

thongdk8:
## Description

This PR adds comprehensive TLS support to the Scalar Manager Helm chart, enabling secure communication for both inbound connections (clients to Scalar Manager) and outbound connections (Scalar Manager to external services). The implementation provides flexible certificate management options, supporting both manual certificate provisioning and automated certificate generation through cert-manager integration.

## Related issues and/or PRs

> No related issues or PRs at this time.

### 1. **New TLS Configuration Structure**
- Added unified TLS configuration under `scalarManager.tls` in values.yaml
- Implemented downstream (inbound) TLS settings for securing client connections
- Added upstream (outbound) TLS configurations for secure connections to:
  - Prometheus (metrics collection)
  - ScalarDB (database connections)
  - ScalarDL (distributed ledger connections)
  - Grafana (dashboard proxy connections)

### 2. **Certificate Management**
- Integrated cert-manager support for automated certificate lifecycle management
- Added support for both CA-based and direct self-signed certificates
- Implemented PKCS#8 private key format for Spring Boot 3 compatibility
- Created new `certmanager.yaml` template for automated certificate generation

### 3. **Deployment Enhancements**
- Updated deployment.yaml with comprehensive environment variables for TLS configuration
- Added volume mounts for TLS certificates and CA certificates
- Implemented conditional volume creation based on TLS settings
- Ensured both API and Web components share the same TLS certificate for simplified management

### 4. **Application Configuration**
- Enhanced applicationProperties with Spring Boot SSL configuration
- Added support for both ECDSA and RSA cipher suites
- Implemented environment variable placeholders for dynamic TLS configuration

### 5. **Documentation Updates**
- Updated README.md with all new TLS configuration parameters
- Enhanced configure-custom-values-scalar-manager.md with comprehensive TLS examples
- Added clear guidance for both development (self-signed) and production (manual certificates) scenarios

## Checklist

- [ ] Tests (unit, integration, etc.) have been added for the changes.

### Key Design Decisions:
1. **Unified TLS Certificate**: Both API and Web components use the same TLS certificate, reducing complexity and certificate management overhead.
2. **Backward Compatibility**: All TLS features are disabled by default, ensuring existing deployments continue to work without modification.
3. **Flexible Certificate Sources**: Support for both manual certificate management (for production) and automated cert-manager integration (for development/testing).
4. **Comprehensive Upstream Support**: Every external service connection can be independently configured for TLS, allowing gradual migration to secure communication.

### Testing Recommendations:
- Test with cert-manager self-signed certificates for development environments
- Verify manual certificate mounting for production scenarios
- Validate upstream TLS connections to each supported service
- Ensure non-TLS deployments continue to function correctly

## Release notes

Added comprehensive TLS support to Scalar Manager Helm chart, including:
- Inbound TLS for securing client connections to Scalar Manager
- Outbound TLS for secure connections to Prometheus, ScalarDB, ScalarDL, and Grafana
- Cert-manager integration for automated certificate management
- Support for both self-signed certificates (development) and manual certificates (production)
- Spring Boot 3 compatible PKCS#8 private key format


charts/scalar-manager/README.md
```diff
@@ -33,6 +33,39 @@ Current chart version is `3.0.0-SNAPSHOT`
...
+| scalarManager.tls.certManager.dnsNames | list | `["localhost","scalar-manager"]` | DNS names for the certificate. |
+| scalarManager.tls.certManager.duration | string | `"8760h0m0s"` | Duration of the certificate. |
+| scalarManager.tls.certManager.enabled | bool | `false` | Enable cert-manager integration. |
+| scalarManager.tls.certManager.issuerRef | object | `{}` | Reference to the issuer for the certificate. |
+| scalarManager.tls.certManager.privateKey | object | `{"algorithm":"RSA","encoding":"PKCS8","size":2048}` | Private key configurations. |
+| scalarManager.tls.certManager.renewBefore | string | `"360h0m0s"` | How long before expiry the certificate should be renewed. |
+| scalarManager.tls.certManager.selfSigned | object | `{"enabled":false,"type":"ca"}` | Use a self-signed certificate. |
+| scalarManager.tls.certManager.selfSigned.type | string | `"ca"` | Type of self-signed issuer (ca or selfSigned). |
```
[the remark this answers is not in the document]
thongdk8: It is genereated so let's leave like that

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -54,15 +111,131 @@ spec:
...
+              - secret:
+                  name: {{ .Values.scalarManager.tls.downstream.privateKeySecret }}
+                  items:
+                    - key: tls.key
+                      path: tls.key
+        {{- end }}
+        {{- if and .Values.scalarManager.tls.downstream.enabled .Values.scalarManager.tls.certManager.enabled }}
+        - name: scalar-manager-tls-volume
+          secret:
+            secretName: {{ include "scalar-manager.fullname" . }}-tls-cert
```
kota2and3kan, line 197:
```suggestion
        - name: scalar-manager-tls-volume
          secret:
            secretName: {{ include "scalar-manager.fullname" . }}-tls-cert
            items:
              - key: tls.key
                path: tls.key
              - key: tls.crt
                path: tls.crt
```
I think cert-manager generates the secret with keys `tls.key` and `tls.crt`. In other words, those key names are guaranteed by cert-manager.

However, I think it might be better to check the key names here for safety, just in case.

It can also keep the consistency between other volume mount configurations in this chart.

What do you think?

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -40,10 +40,67 @@ spec:
           imagePullPolicy: {{ .Values.scalarManager.api.image.pullPolicy }}
           securityContext:
             {{- toYaml .Values.scalarManager.securityContext | nindent 12 }}
+          env:
```
kota2and3kan, line 43: I think we need to configure Scalar Manager Web via environment variables because there is no configuration files like `application.properties`.

However, I think we can configure Scalar Manager API by using `application.properties`. And, it seems that there are two ways to configure Scalar Manager API in this update. One is using the `application.properties` file. Another is via environment variables.

My concern is that mixing two configuration ways might cause confusion for users. Is it possible to use only one of `application.properties` or environment variables?

Also, we already provide the way to configure Scalar Manager API by using `application.properties`. Therefore, to keep backward compatibility and consistency of `How to configure Scalar Manager API`, I think it would be better to use only `application.properties` instead of mixing both `application.properties` and environment variables.

What do you think?

charts/scalar-manager/values.yaml
```diff
@@ -121,6 +122,40 @@ scalarManager:
...
+      # ============================================
+      # TLS Configuration
+      # ============================================
+
+      # Spring Boot SSL configuration will be inferred from our custom tls inbound configurations
+      server.ssl.enabled=${app.tls.inbound.enabled}
+      server.ssl.certificate=${app.tls.inbound.cert-path}
+      server.ssl.certificate-private-key=${app.tls.inbound.private-key-path}
+      # Include both ECDSA and RSA ciphers for maximum compatibility
+      server.ssl.ciphers=TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
```
kota2and3kan, line 134: Do we need to set the ciphers configuration in the `application.properties` explicitly?

I think it would be better to use `default` as long as possible if there is no problem.

This is because the `default` (recommended ciphers) might be updated dynamically by following the trend or some updates on the library side.

My concern is that setting the concrete value might cause `Users use the old (non-recommended) cipher in the future version`.

Therefore, I think it would be better to use the `default` (default of library or flamework) if there is no strong reason to set some concrete values.

What do you think?

kota2and3kan (commented):
Thank you for the updates!
I left several comments.
Please take a look when you have time! 🙇‍♂️

charts/scalar-manager/values.yaml
```diff
@@ -160,4 +195,87 @@ scalarManager:
...
+      dnsNames:
+        - localhost
+        - scalar-manager
+      # -- Reference to the issuer for the certificate.
+      issuerRef: {}
+      # -- Use a self-signed certificate.
+      selfSigned:
+        enabled: false
+        # -- Type of self-signed issuer (ca or selfSigned).
+        type: ca
```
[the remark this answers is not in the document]
thongdk8: These options will be available if we want the certs for Scalar Manager to be managed by itself (when the `certManager.selfSigned.enabled: true`
For the details, I summarized as below:

Two Self-Signed Types:
  1. ca type - Creates a self-signed Certificate Authority (CA) chain
  2. selfSigned type - Creates a direct self-signed certificate

For the CA Type (type: ca):
  - Creates a self-signed CA certificate first
  - Uses that CA to sign the actual TLS certificate
  - Results in a proper certificate chain (CA → Server Certificate)
  - The CA certificate can be distributed as a trust anchor

For the SelfSigned Type (type: selfSigned):
  - Creates a single self-signed certificate directly
  - No certificate chain - the certificate signs itself
  - The certificate itself serves as the trust anchor

Use Cases:
  When to use ca:
  - Production-like environments where you want to simulate proper PKI hierarchy
  - Multiple services that need certificates from the same CA
  - Client trust distribution - easier to distribute one CA cert to clients
  - Future flexibility - can issue multiple certificates from the same CA

  When to use selfSigned:
  - Quick testing where certificate chain doesn't matter
  - Single service deployments
  - Simplified setup with fewer certificate resources

Both options are valuable because they serve different deployment scenarios. The ca type provides better security practices, while selfSigned offers simplicity for development/testing.
The choice depends on the requirements for certificate management I think.
WDYT?

thongdk8 (commented):
@kota2and3kan Thank you for checking the PR, I made some responses to your comments. PTAL when you have time

charts/scalar-manager/values.yaml
```diff
@@ -160,4 +195,87 @@ scalarManager:
...
+      prometheus:
+        # -- Enable TLS for Prometheus connections.
+        enabled: false
+        # -- Secret containing the CA root certificate for validating Prometheus SSL certificates.
+        caRootCertSecret: ""
+        # -- Override the hostname for TLS SNI when connecting to Prometheus.
+        overrideAuthority: ""
+
+      # -- ScalarDB TLS configuration.
+      scalardb:
```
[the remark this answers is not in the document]
kota2and3kan: I discussed with @feeblefakie  and @ypeckstadt  in the meeting. As a result, we decided to update Scalar Manager.

- https://scalar-labs.slack.com/archives/C05UBRUH33N/p1753412583309169?thread_ts=1753412274.602559&cid=C05UBRUH33N

You may already know it, but as mentioned above, first, we will update the Scalar Manager side. After that, let's discuss the updates of the Helm Chart based on the new Scalar Manager.

charts/scalar-manager/templates/scalar-manager/deployment.yaml
kota2and3kan: I have one question.

If we don't set `app.tls.inbound.enabled=${APP_TLS_INBOUND_ENABLED}` in the `application.properties` file explicitly as follows, and we set `scalarManager.tls.downstream.enabled=true`, does Scalar Manager API start with TLS enabled configuration?

```yaml
scalarManager:
  api:
    applicationProperties: |
      spring.datasource.url=jdbc:postgresql://user.postgresql.example.com:5432/postgres
      spring.datasource.username=postgres
      spring.datasource.password=postgres
      spring.datasource.driver-class-name=org.postgresql.Driver
  tls:
    downstream:
      enabled: true
```

In this case, I think some default configurations like `app.tls.inbound.enabled=${APP_TLS_INBOUND_ENABLED}` are not applied because the default `application.properties` is overwritten by the above user-specified `application.properties`. Is my understanding correct?

If my understanding is correct, Scalar Manager API does not start with TLS configuration even if we set `scalarManager.tls.downstream.enabled=true` because there is no configuration about TLS on the overwritten (user-specified) `application.properties` side.

That's why I want to ask this question. I'd like to know the behavior of the Scalar Manager API in the above configuration.

charts/scalar-manager/values.yaml
kota2and3kan: Ah, sorry. I lacked descriptions.

I agree with making the acceptable cipher suites configurable.

However, I think we should not set the default (hard-coded default) value for it on our side.

Regarding the default value of `server.ssl.ciphers`, I think we should use the library's default as a default value so that we can follow the security trend.

For example, set the default property (default values.yaml) as follows:

```yaml
scalarManager:
  api:
    applicationProperties: |
      (omit other configurations)
      
      # By default, Scalar Manager API accepts the library default cipher suites. You can configure any cipher suites you prefer if you want.
      server.ssl.ciphers=
```

If users do not specify any cipher suites, the library's default (based on the security trend) is set.

If users want to set their preferred cipher suites for some reason, users can overwrite them in the `application.properties` file.

This can keep the default value secure based on the security trend (library's default), and this can provide flexible options for users.

As above, my suggestions are the following:

- Make the acceptable cipher suites configurable (As you already did).
- However, don't set the default value on our side (i.e., set empty or some special value to use the library's default value). In other words, removing the current default values that are set on the Helm Chart (`values.yaml` file) side.

kota2and3kan:
@thongdk8 
Thank you for your answers!
As we discussed in Slack and the meeting, I think we need additional discussion and update Scalar Manager itself first, to support monitoring several Scalar product deployments.
However, I left some comments based on the current state, just in case.
Please take a look when you have time! 🙇‍♂️

charts/scalar-manager/values.yaml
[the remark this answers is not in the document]
thongdk8: Thank you for your comments and suggestions. I agree with that, doing it will simplify our chart. I made updates to apply the suggestions, PTAL again when you have time.

charts/scalar-manager/values.yaml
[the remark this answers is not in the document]
thongdk8: @kota2and3kan I have made some updates to support multiple namespaces. PTAL again when you have time

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -39,10 +39,105 @@ spec:
...
+            # Spring Boot SSL configuration (inbound TLS)
+            - name: SERVER_SSL_ENABLED
+              value: "{{ .Values.scalarManager.tls.downstream.enabled }}"
+            {{- if .Values.scalarManager.tls.downstream.enabled }}
+            - name: SERVER_SSL_CERTIFICATE
+              value: "/tls/scalar-manager/certs/tls.crt"
+            - name: SERVER_SSL_CERTIFICATE_PRIVATE_KEY
+              value: "/tls/scalar-manager/certs/tls.key"
+            {{- end }}
+            # Upstream TLS configuration
```
kota2and3kan:
```suggestion
            # Upstream TLS configuration
            # Note: This upstream TLS feature is now in preview. We should test more and update if necessary in the future.
```

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -39,10 +39,105 @@ spec:
...
+            {{- end }}
           volumeMounts:
             - name: api-application-properties-volume
               mountPath: /app/application.properties
               subPath: scalar-manager-api-application.properties
+            {{- if .Values.scalarManager.tls.downstream.enabled }}
+            - name: scalar-manager-tls-volume
+              mountPath: /tls/scalar-manager/certs
+            {{- end }}
+            {{- range $index, $prometheus := .Values.scalarManager.tls.upstream.prometheus }}
```
kota2and3kan:
```suggestion
            # Note: This upstream TLS feature is now in preview. We should test more and update if necessary in the future.
            {{- range $index, $prometheus := .Values.scalarManager.tls.upstream.prometheus }}
```

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -53,15 +148,157 @@ spec:
...
+              value: "/tls/ca/certs/ca.crt"
+              {{- else if .Values.scalarManager.tls.certManager.enabled }}
+              # For cert-manager issuers (self-signed or external), we assume ca.crt is present in the secret.
+              value: "/tls/scalar-manager/certs/ca.crt"
+              {{- else }}
+              # Manual mode without a caRootCertSecret, assume the server cert is the trust anchor.
+              value: "/tls/scalar-manager/certs/tls.crt"
+              {{- end }}
+            {{- end }}
+            {{- range $index, $grafana := .Values.scalarManager.tls.upstream.grafana }}
```
kota2and3kan:
```suggestion
            # Note: This upstream TLS feature is now in preview. We should test more and update if necessary in the future.
            {{- range $index, $grafana := .Values.scalarManager.tls.upstream.grafana }}
```

charts/scalar-manager/templates/scalar-manager/deployment.yaml
```diff
@@ -53,15 +148,157 @@ spec:
...
+        {{- end }}
+        {{- if .Values.scalarManager.tls.downstream.caRootCertSecret }}
+        - name: scalar-manager-tls-downstream-ca-volume
+          secret:
+            secretName: {{ .Values.scalarManager.tls.downstream.caRootCertSecret }}
+            items:
+              - key: ca.crt
+                path: ca.crt
+        {{- end }}
+        {{- range $index, $grafana := .Values.scalarManager.tls.upstream.grafana }}
```
kota2and3kan:
```suggestion
        # Note: This upstream TLS feature is now in preview. We should test more and update if necessary in the future.
        {{- range $index, $grafana := .Values.scalarManager.tls.upstream.grafana }}
```

charts/scalar-manager/values.yaml
```diff
@@ -163,3 +189,104 @@ scalarManager:
...
+      enabled: false
+      # -- Certificate configurations when cert-manager is not enabled.
+      # -- Secret containing the certificate for downstream TLS.
+      certChainSecret: ""
+      # -- Secret containing the private key for downstream TLS.
+      privateKeySecret: ""
+      # -- Secret containing the CA root certificate for web-to-API communication (Web needs to validate API's TLS certificate).
+      caRootCertSecret: ""
+
+    # -- Upstream TLS configuration for external service connections.
```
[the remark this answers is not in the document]
thongdk8: Applied in [e15ec97](https://github.com/scalar-labs/helm-charts/pull/303/commits/e15ec974326b26b0a6565b53f93566ecb0ddfee6)

charts/scalar-manager/templates/scalar-manager/deployment.yaml
thongdk8: Thank you for checking and leaving comments. I have added the notes as you suggested in [e15ec97](https://github.com/scalar-labs/helm-charts/pull/303/commits/e15ec974326b26b0a6565b53f93566ecb0ddfee6)

kota2and3kan (approved):
LGTM! Thank you!

ypeckstadt (approved):
LGTM. Thank you.

